PR #50, opened 2022-12-02
Fix: Add Transactions Button

Yash-Chaubal:
Add Tx btn now works even after txs are submitted. Also added a fix for multicalls where after the main batch had been signed, we got more signature requests for individual steps - therefore potentially triggering a tx twice.


src/providers/TransactionsDispatcherContextProvider.tsx
```diff
@@ -153,54 +195,67 @@ const TransactionsDispatcherContextProvider = ({ children }: { children: ReactNo
...
+    const updatedCrossChainActions = crossChainActions.map(
+      (crossChainAction) => {
+        if (
+          crossChainAction.id !== firstUnsentCrossChainAction.id &&
+          (!crossChainAction.multiCallData?.id ||
+            firstUnsentCrossChainAction.multiCallData?.id !==
+              crossChainAction.multiCallData.id)
+        ) {
+          return crossChainAction;
+        }
```
Yash-Chaubal, line 224: @poocart Basically changed the `||` back to `&&` but also added another check - the reason for this is that in a multicall, even after the tx had been signed, it still made us sign the subsequent steps, trigger the tx twice.





[3:40](https://pillarwallet.slack.com/archives/C04DD82G8G5/p1669995615580619)
This check makes sure that the current action to be executed is not part of a multicall that's already been executed

IAmKio (approved):

poocart (approved):
LGTM! 🔥 also nice job on cleanup, but would be good to not include more cleanup than the feature code added because then it's hard to understand what is new haha, thanks! 🙏

src/providers/TransactionsDispatcherContextProvider.tsx
poocart: Thanks

IAmKio:
@poocart do you know if this is still relevent?

poocart:
> @poocart do you know if this is still relevent?

I don't think so, closed and parked for now.

PR #63, opened 2022-12-22
Smart wallet balance against each chain

MayukhSamadder2000:
Shown asset balance in usd against each chain for Smart wallet
image

PR #73, opened 2023-01-09
Balance against keybased networks

MayukhSamadder2000:
Screenshot 2023-01-09 at 8 27 10 PM


poocart:
I'm sorry Mayukh, but this PR is mostly changes of https://github.com/etherspot/etherspot-react-transaction-buidler/pull/63/files, maybe you want to change merge target to that branch and merge to it first and then merge result to develop?

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -320,38 +325,97 @@ const NetworkAssetSelectInput = ({
...
+        return;
+
+      setKeybasedWalletBalanceByChain([
+        ...keybasedWalletBalanceByChain.filter(
+          (element) => element.chain !== CHAIN_ID.AVALANCHE
+        ),
+        {
+          total: filteredSelectedNetworkAssets.reduce((acc, curr) => {
+            return curr.balanceWorthUsd ? acc + curr.balanceWorthUsd : acc;
+          }, 0),
```
Yash-Chaubal, line 406: Could probably shorten this, and I added `?` just to be safe. Also could you change `curr` to a more descriptive name, like `asset` or something

```suggestion
          total: filteredSelectedNetworkAssets.reduce((acc, curr) => curr?.balanceWorthUsd ? acc + curr.balanceWorthUsd : acc, 0),
```

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -377,12 +441,21 @@ const NetworkAssetSelectInput = ({
...
         : '$0';
-    } else if (accType === DestinationWalletEnum.Key && label === 'From') {
-      return '$0';
+    } else if (
+      accType === DestinationWalletEnum.Key &&
+      label === 'From' &&
+      keybasedWalletBalanceByChain?.length
+    ) {
+      let balanceByChain = keybasedWalletBalanceByChain.filter(
+        (item: any) => item.chain === supportedChain.chainId
```
Yash-Chaubal, line 452: Is there a way to avoid using `any`? Does `keybasedWalletBalanceByChain` have a type or something? Isn't to important in this case, but better to avoid it where you can. Maybe change this to `item?.chain` since we're using `any`

src/utils/common.ts
```diff
@@ -74,8 +74,8 @@ export const getTimeBasedUniqueId = (): string => uniqueId(`${+new Date()}-`);
 
 export const formatMaxAmount = (maxAmountBN: BigNumber, decimals: number): string => ethers.utils.formatUnits(maxAmountBN, decimals);
 
-export const sumAssetsBalanceWorth = (supportedAssets: IAssetWithBalance[]) => {
-  return supportedAssets.reduce((sum, asset) => {
+export const sumAssetsBalanceWorth = (supportedAssets: any) => {
+  return supportedAssets.reduce((sum: any, asset: { balanceWorthUsd: any; }) => {
```
Yash-Chaubal, line 78: Same thing as above with `any` and also in a previous reducer you used `acc` instead of sum, could we stay consistent with the naming

Yash-Chaubal (changes requested):
Looking good! Just a few housekeeping comments

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -320,38 +325,97 @@ const NetworkAssetSelectInput = ({
...
-    walletAddress,
+    accountAddress,
+    preselectedNetwork,
+    filteredSelectedNetworkAssets,
+    isLoadingAssets,
+  ]);
+
+  useEffect(() => {
+    console.log("smartWalletBalanceByChain",smartWalletBalanceByChain)
+  }, [smartWalletBalanceByChain])
```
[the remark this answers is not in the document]
MayukhSamadder2000: No there's no need thank's for pointing it out.

src/utils/common.ts
MayukhSamadder2000: yeah used asset everywhere

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -320,38 +325,92 @@ const NetworkAssetSelectInput = ({
       await getSmartWalletBalancesByChain(walletAddress, supportedChains);
     };
     handleBalanceGet();
-  }, [supportedChains, sdk, walletAddress]);
+  }, [supportedChains, sdk, accountAddress]);
+
+  useEffect(() => {
+    const handleKeybasedBalanceGet = async () => {
+      if (!sdk || !providerAddress) return;
+      await getKeybasedWalletBalancesPerChain(providerAddress, supportedChains);
```
poocart, line 333: I'm thinking if these shouldn't go somwhere in the upper level like `EtherspotContextProvider` because everytime we will need balances we have to add these lines, right? wdyt

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -320,38 +325,92 @@ const NetworkAssetSelectInput = ({
...
+        !filteredSelectedNetworkAssets.length ||
+        isLoadingAssets
       ) {
         return;
       }
       setsmartWalletBalanceByChain([
-        ...smartWalletBalanceByChain,
+        ...smartWalletBalanceByChain.filter(
+          (element) => element.chain !== CHAIN_ID.AVALANCHE
+        ),
```
poocart, line 358: Why not pulling current state from `setsmartWalletBalanceByChain`?

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -320,38 +325,92 @@ const NetworkAssetSelectInput = ({
...
     supportedChains,
-    walletAddress,
+    accountAddress,
+    preselectedNetwork,
+    filteredSelectedNetworkAssets,
+    isLoadingAssets,
+  ]);
+
+  useEffect(() => {
+    const updateAvalancheBalanceKeybased = () => {
```
poocart, line 381: Same as for other smart wallet / key based balance methods: maybe we can move all and combine? some code can definitely be reused and every time we would need total balances we now would need to move this code into other components which would be wrong

poocart: This way we would also be able to update the balance using hook (if method exposed)

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -377,12 +436,21 @@ const NetworkAssetSelectInput = ({
         (item: any) => item.chain === supportedChain.chainId
       );
       return smartWalletBalanceByChain?.length && balanceByChain.length
-        ? formatAmountDisplay(String(balanceByChain[0].total), '$')
+        ? ` · ${formatAmountDisplay(String(balanceByChain[0].total), '$')}`
         : '$0';
-    } else if (accType === DestinationWalletEnum.Key && label === 'From') {
-      return '$0';
+    } else if (
```
poocart, line 441: No need `else if` as previous condition returns the output so previous condition can be just closed and new put new `if`

src/providers/EtherspotContextProvider.tsx
```diff
@@ -316,6 +317,42 @@ const EtherspotContextProvider = ({
...
+          .filter((element) => element.chainId !== CHAIN_ID.AVALANCHE)
+          .map(async (element) => {
+            try {
+              let supportedAssets =
+                await getSupportedAssetsWithBalancesForChainId(
+                  element.chainId,
+                  true,
+                  walletAddress,
+                  false
+                );
```
poocart, line 335:
```suggestion
              let supportedAssets = await getSupportedAssetsWithBalancesForChainId(
                element.chainId,
                true,
                walletAddress,
                false
              );
```

src/providers/EtherspotContextProvider.tsx
```diff
@@ -316,6 +317,42 @@ const EtherspotContextProvider = ({
...
+              balanceByChain.push({
+                title: element.title,
+                chain: element.chainId,
+                total: supportedAssets.reduce((sum, asset) => {
```
poocart, line 339: Reuse your sum method

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -320,38 +325,92 @@ const NetworkAssetSelectInput = ({
...
+        isLoadingAssets
+      )
+        return;
+
+      setKeybasedWalletBalanceByChain([
+        ...keybasedWalletBalanceByChain.filter(
+          (element) => element.chain !== CHAIN_ID.AVALANCHE
+        ),
+        {
+          total: filteredSelectedNetworkAssets.reduce((sum, asset) => {
```
poocart, line 399: reuse your `sumAssetsBalanceWorth` method

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
MayukhSamadder2000: Do you mean  `EtherspotContextProvider` or `TransactionBuilderContextProvider`?

poocart: `EtherspotContextProvider` because it holds values for accounts generally

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -320,38 +325,84 @@ const NetworkAssetSelectInput = ({
...
+  ]);
+
+  useEffect(() => {
+    const updateAvalancheKeybasedBalance = () => {
+      if (
+        !sdk ||
+        !walletAddress ||
+        !preselectedNetwork ||
+        preselectedNetwork.chainId !== CHAIN_ID.AVALANCHE ||
+        !keybasedWalletBalanceByChain ||
```
poocart, line 383: Is this still needed when u added `prev` on `setKeybasedWalletBalanceByChain`?

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -320,38 +325,84 @@ const NetworkAssetSelectInput = ({
...
   }, [selectedNetworkAssets, assetSearchQuery]);
 
   useEffect(() => {
-    const updateAvalancheBalance = () => {
+    const updateAvalancheSmartWalletBalance = () => {
       if (!sdk ||
         !walletAddress ||
         !preselectedNetwork ||
         preselectedNetwork.chainId !== CHAIN_ID.AVALANCHE ||
         !smartWalletBalanceByChain ||
```
poocart, line 349: Is this still needed when u added `prev` on `setSmartWalletBalanceByChain `?

src/providers/EtherspotContextProvider.tsx
```diff
@@ -311,11 +312,41 @@ const EtherspotContextProvider = ({
...
+        supportedChains
+          .filter((element) => element.chainId !== CHAIN_ID.AVALANCHE)
+          .map(async (element) => {
+            try {
+              let supportedAssets = await getSupportedAssetsWithBalancesForChainId(
+                  element.chainId,
+                  true,
+                  walletAddress,
+                  false
+                );
```
poocart, line 334:
```suggestion
              let supportedAssets = await getSupportedAssetsWithBalancesForChainId(
                element.chainId,
                true,
                walletAddress,
                false
              );
```

src/utils/common.ts
```diff
@@ -74,8 +74,8 @@ export const getTimeBasedUniqueId = (): string => uniqueId(`${+new Date()}-`);
 
 export const formatMaxAmount = (maxAmountBN: BigNumber, decimals: number): string => ethers.utils.formatUnits(maxAmountBN, decimals);
 
-export const sumAssetsBalanceWorth = (supportedAssets: IAssetWithBalance[]) => {
-  return supportedAssets.reduce((sum, asset) => {
+export const sumAssetsBalanceWorth = (supportedAssets: any) => {
+  return supportedAssets.reduce((sum: number, asset: { balanceWorthUsd: number; }) => {
```
poocart, line 78: `asset` can stay as `IAssetWithBalance`

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -314,44 +319,68 @@ const NetworkAssetSelectInput = ({
...
-    updateAvalancheBalance();
+    updateAvalancheSmartWalletBalance();
   }, [
     sdk,
     supportedChains,
-    walletAddress,
+    accountAddress,
+    preselectedNetwork,
+    filteredSelectedNetworkAssets,
+    isLoadingAssets,
```
poocart, line 354: hmmm adding `isLoadingAsset` might fall into loop, hope you bullet tested this

src/utils/common.ts
```diff
@@ -75,7 +75,7 @@ export const getTimeBasedUniqueId = (): string => uniqueId(`${+new Date()}-`);
 export const formatMaxAmount = (maxAmountBN: BigNumber, decimals: number): string => ethers.utils.formatUnits(maxAmountBN, decimals);
 
 export const sumAssetsBalanceWorth = (supportedAssets: IAssetWithBalance[]) => {
-  return supportedAssets.reduce((sum, asset) => {
+  return supportedAssets.reduce((sum: number, asset: IAssetWithBalance ) => {
```
poocart, line 78:
```suggestion
  return supportedAssets.reduce((sum: number, asset: IAssetWithBalance) => {
```

src/providers/EtherspotContextProvider.tsx
```diff
@@ -293,14 +295,44 @@ const EtherspotContextProvider = ({
...
+  const getKeybasedWalletBalancesPerChain = useCallback(
+    async (walletAddress: string, supportedChains: Chain[]) => {
+      if (!sdk || !walletAddress) return;
+      let balanceByChain: IBalanceByChain[] = [];
+      await Promise.all(
+        supportedChains
+          .filter((element) => element.chainId !== CHAIN_ID.AVALANCHE)
+          .map(async (element) => {
+            try {
+              let supportedAssets = await getSupportedAssetsWithBalancesForChainId(
```
poocart, line 330: this might cause API rate limit to block u, solutions are either put some cooldowns or make it sequential (might cause UX problems), but if you checked and it's ok then all good

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
```diff
@@ -314,44 +319,68 @@ const NetworkAssetSelectInput = ({
...
+  ]);
+
+  useEffect(() => {
+    const updateAvalancheKeybasedBalance = () => {
+      if (!walletAddress ||
+        !preselectedNetwork ||
+        preselectedNetwork.chainId !== CHAIN_ID.AVALANCHE ||
+        !filteredSelectedNetworkAssets.length ||
+        accountType !== DestinationWalletEnum.Key ||
+        isLoadingAssets
```
poocart, line 364: same as per comment in the above

src/components/NetworkAssetSelectInput/NetworkAssetSelectInput.tsx
MayukhSamadder2000: yeah since the function is keeping a check for isLoadingAssets

src/providers/EtherspotContextProvider.tsx
MayukhSamadder2000: you mean some sort of throttling?

Yash-Chaubal (approved):

PR #74, opened 2023-01-10
PRO-965 Klima Dao Bug Fixes

vignesha22:
## Description

- Added fromAddress to klima get cross chain quotes method which gets offer based on the source address
- Removed chainId check for key based wallet execution since chainChain fn doesn't return error while the source chainId and the destination chainId are same

## Motivation and Context

-

## How Has This Been Tested?

- Tested Locally
-
-

## Types of changes

### DIFF
--- a/src/utils/transaction.ts
+++ b/src/utils/transaction.ts
@@ -770,10 +770,9 @@ export const submitWeb3ProviderTransactions = async (
   }
 
   // TODO: check against current
-  if (chainId !== 1) {
-    const changed = await changeToChain(chainId);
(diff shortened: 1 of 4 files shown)


poocart (approved):
lgtm

src/utils/transaction.ts
```diff
@@ -770,10 +770,9 @@ export const submitWeb3ProviderTransactions = async (
   }
 
   // TODO: check against current
```
poocart, line 772: This can be removed as well per changes below
```suggestion
```

PR #77, opened 2023-01-19
PRO-967 - Display Gas Price

vignesha22:
## Description

- Display gas price from offer if the estimation fails when its from key based account for Klima Block only

## Motivation and Context

-

## How Has This Been Tested?

- Tested Locally
-
-

## Types of changes

### DIFF
--- a/src/components/TransactionPreview/ActionPreview.tsx
+++ b/src/components/TransactionPreview/ActionPreview.tsx
@@ -462,7 +462,15 @@ const ActionPreview = ({
 
   const cost = useMemo(() => {
     if (isEstimating) return "Estimating...";
-    if (!estimated || !estimated?.gasCost) return estimated?.errorMessage;
+    if (!estimated || !estimated?.gasCost) {
+      if (crossChainAction.type === TRANSACTION_BLOCK_TYPE.KLIMA_STAKE && crossChainAction.useWeb3Provider && crossChainAction.gasCost) {
+        return formatAmountDisplay(
+          `${crossChainAction.gasCost}`,
+          "$"
+        );
+      }
+      return estimated?.errorMessage;
+    }
 
     const gasCostNumericString = estimated.feeAmount && crossChainAction.gasTokenDecimals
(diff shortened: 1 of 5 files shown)


src/types/crossChainAction.ts
```diff
@@ -73,6 +73,7 @@ interface KlimaStakingAction {
 	containsSwitchChain?: boolean;
 	receiveAmount?: string;
 	bridgeUsed?: string;
+  gasCost?: string;
```
poocart, line 76: hmmm, why you're not using `ICrossChainAction` `estimated` for `gasCost`?

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -462,7 +462,15 @@ const ActionPreview = ({
 
   const cost = useMemo(() => {
     if (isEstimating) return "Estimating...";
-    if (!estimated || !estimated?.gasCost) return estimated?.errorMessage;
+    if (!estimated || !estimated?.gasCost) {
+      if (crossChainAction.type === TRANSACTION_BLOCK_TYPE.KLIMA_STAKE && crossChainAction.useWeb3Provider && crossChainAction.gasCost) {
+        return formatAmountDisplay(
+          `${crossChainAction.gasCost}`,
+          "$"
+        );
```
poocart, line 470: can stay in one line it's fine, single quotes please, no need to wrap as string for `gasCost`

```suggestion
        return formatAmountDisplay(crossChainAction.gasCost, '$');
```

src/types/crossChainAction.ts
vignesha22: that's because the estimated object has two other properties and the gasCost in estimated is in terms of Ethers but this is straight USD cost from offer

src/components/TransactionPreview/ActionPreview.tsx
vignesha22: right got it

PR #84, opened 2023-01-27
PLR DAO Membership Block [PRO-878]

parag-chauhan:
### BUIDLER: PLR DAO Membership Block
- Transfer their assets crosschain to the PLR DAO Contract chain (should it need to be transferred)
- Stake the PLR tokens into the PLR DAO smart contract

## Motivation and Context
This block is aimed at making the process of joining the PLR DAO easy via the BUIDLER component with the implementation of a PLR  DAO Membership Block. This will allow a user to choose their PLR tokens that they may own on any chain, and ultimately stake it into the PLR DAO staking contract and become a member.

## How Has This Been Tested?
- Tested on Polygoan Mainnet
  - If User have enoughPLR Token on Polygon then user sees the Staking screen
  - If User doen't have enoughPLR Token on Polygon then user sees the Swap screen

#### Main Screen
image

#### Swap
image

#### Staking (e.g: `MAX_PLR_TOKEN_LIMIT` = 5)
image

## Types of changes

- [x] New feature (non-breaking change which adds functionality)


src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  fromAsset?: IAssetWithBalance;
+  selectedAsset?: IAssetWithBalance | null;
+  toAsset?: TokenListToken;
+  fromAssetDecimals?: number;
+  fromAssetSymbol?: string;
+  amount?: string;
+  accountType: AccountTypes;
+  receiverAddress?: string;
+  offer?: ExchangeOffer;
+}
```
IAmKio, line 51: Could we re-organise and split these imports into sections pleasse

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  flex-direction: row;
+  align-items: center;
+  font-family: 'PTRootUIWebMedium', sans-serif;
+`;
+
+const Container = styled.div`
+  align-items: center;
+  font-family: 'PTRootUIWebMedium', sans-serif;
+  background: #21002e;
+  color: #fefefe;
```
IAmKio, line 77: I understand that this has been lifted from the theme but we need to try and accmmodate the custom theming props. Have a chat with Max if you're struggling.

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+const mapOfferToOption = (offer: ExchangeOffer) => {
+  const serviceDetails = swapServiceIdToDetails[offer.provider];
+  return {
+    title: serviceDetails?.title ?? offer.provider,
+    value: offer.provider,
+    iconUrl: serviceDetails?.iconUrl,
+  };
+};
+
+// TODO:UNCOMMENT
```
IAmKio, line 127: Please remove any unneeded comment blocks

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+
+  const getWalletBalance = async (chainId: number, name: string) => {
+    const accountBalancePromise = [];
+    if (accountAddress && providerAddress) {
+      accountBalancePromise.push(getSupportedAssetsWithBalancesForChainId(chainId, true, accountAddress));
+      accountBalancePromise.push(getSupportedAssetsWithBalancesForChainId(chainId, true, providerAddress));
+    }
+
+    return await Promise.allSettled(accountBalancePromise)
+      .then((response: any[]) => {
```
IAmKio, line 223: It may be worth checking if each `allSettled` response contains an error or not what you're looking for

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+    chainPromise.push(getWalletBalance(supportedChains[1].chainId, supportedChains[1].title))
+    // chainPromise.push(getWalletBalance(supportedChains[2].chainId, supportedChains[2].title))
+    chainPromise.push(getWalletBalance(supportedChains[3].chainId, supportedChains[3].title))
+    // chainPromise.push(getWalletBalance(supportedChains[3].chainId, supportedChains[3].title))
+    // supportedChains.forEach(async (chain) => {
+    //   chainPromise.push(getWalletBalance(chain.chainId, chain.title))
+    // });
+    return await Promise.allSettled(chainPromise).catch((e) => {
+      return [];
+    });
```
IAmKio, line 278: Please remember to remove dead code

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  }
+
+  const fetchAllAccountBalances = async () => {
+    var accountBalanceWithSupportedChains = await getBalanceForAllChains().then((data) => data.map((d) => d.value));
```
IAmKio, line 282: Remember to `.catch` here just incase

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+          fromAmount: ethers.utils.parseUnits(
+            amount,
+            selectedFromAsset.decimals
+          ),
+          toTokenAddress: hasEnoughPLR ? plrDaoMemberNFT.address : plrDaoAsset.address,
+          fromTokenAddress: selectedFromAsset.address,
+        });
+        return offers;
+      } catch (e) {
+        //
```
IAmKio, line 349: Should something happen here if any of the above fail?

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+      }
+    }, 200),
+    [sdk, selectedFromAsset, amount, selectedFromNetwork, accountAddress, selectedAccountType]
+  );
+
+  useEffect(() => {
+    accountAddress || providerAddress && sdk?.getNftList({
+      account: accountAddress || providerAddress,
+    }).then((output:NftList) => {
+      if(output?.items?.length) {
```
IAmKio, line 359: This endpoint can be error prone like the others, it's worth checking you have a valid response here before determining items length etc

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+    });
+  }, [
+    selectedFromNetwork,
+    selectedFromAsset,
+    amount,
+    selectedOffer,
+    availableOffers,
+    selectedAccountType,
+    receiverAddress,
+  ]);
```
IAmKio, line 440: There are a ton of `useEffect` blocks here. Whilst it's not an anti-pattern per se, please add a comment block explaining what each one does as it's clear you've grouped them.

src/providers/TransactionBuilderContextProvider.tsx
```diff
@@ -644,7 +650,117 @@ const TransactionBuilderContextProvider = ({
...
+        return;
+      }
+
+      const estimated = await estimateCrossChainAction(
+        getSdkForChainId(plrDaoAsset.chainId),
+        web3Provider,
+        crossChainAction.destinationCrossChainAction[0],
+        providerAddress,
+        accountAddress,
+      );
```
IAmKio, line 735: We should add error checking here, as any of these `await`'s can fail, or you can check the results further down the code

src/utils/chain.ts
```diff
@@ -15,6 +15,7 @@ export const CHAIN_ID = {
   FANTOM: 250,
   CELO: 42220,
   MOONBEAM: 1284,
+  GNOSIS:100
```
IAmKio, line 18:
```suggestion
  GNOSIS: 100
```

src/utils/transaction.ts
```diff
@@ -160,6 +160,113 @@ export const klimaDaoStaking = async (
   }
 };
 
+export const plrDaoStaking = async (
```
IAmKio, line 163: Snack case please

```suggestion
export const PlrDaoStaking = async (
```

src/utils/transaction.ts
```diff
@@ -160,6 +160,113 @@ export const klimaDaoStaking = async (
...
+      data: plrDaoStakeTransactionRequest.data,
+      chainId: plrDaoAsset.chainId,
+      value: 0,
+      createTimestamp,
+      status: CROSS_CHAIN_ACTION_STATUS.UNSENT,
+    };
+
+    transactions = [...transactions, plrDaoApprovalTransaction, plrDaoStakinglTransaction];
+
+    if (receiverAddress && receiverAddress != sdk.state.accountAddress) {
```
IAmKio, line 241: Does TS handle `!=` as opposed to `!==`?

IAmKio: FYI: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality

src/utils/transaction.ts
```diff
@@ -314,6 +421,308 @@ export const buildCrossChainAction = async (
...
+            return { errorMessage: 'Failed to get bridge route!' };
+          }
+  
+      } catch (e) {
+        return { errorMessage: 'Failed to get PLR Dao staking transaction!' };
+      }
+    }
+    
+  }
+
```
IAmKio, line 725: Ooooff, this is one huge block of nested if / elses. Could you try and split this up into smaller chunks of code that could possibly be reused? As it looks like some code is repeating itself...

IAmKio (changes requested):
Excellent work in getting this flow up and running! 🎉 
Have a look at the comments below and also please check your linting configuration. @Yash-Chaubal can assist you with this.

src/components/AccountSwitchInput/AccountSwitchInput.tsx
```diff
@@ -27,11 +31,11 @@ const AccountSwitchInput = ({
   showTotals = false,
 }: AccountSwitchInputProps) => {
   let walletOptions = [
-    { title: 'Smart Wallet', value: DestinationWalletEnum.Contract},
+    { title: smartWalletTotal ? `${smartWalletTotal} Smart Wallet`: 'Smart Wallet', value: DestinationWalletEnum.Contract},
```
MayukhSamadder2000, line 34: Hey So I handled the total display through this ticket https://linear.app/pillarproject/issue/PRO-727

src/components/AccountSwitchInput/AccountSwitchInput.tsx
```diff
@@ -27,11 +31,11 @@ const AccountSwitchInput = ({
...
   ];
 
   if (!hideKeyBased) {
-    walletOptions = [{ title: 'Key based', value: DestinationWalletEnum.Key }, ...walletOptions];
+    walletOptions = [{ title: keyBasedWalletTotal ? `${keyBasedWalletTotal} Key based`:'Key based', value: DestinationWalletEnum.Key }, ...walletOptions];
```
MayukhSamadder2000, line 38: same as above

src/components/Button/closeButtonWhite.tsx
```diff
@@ -0,0 +1,28 @@
...
+  top: -20px;
+  position: relative;
+
+  &:hover {
+    opacity: 0.5;
+  }
+  
+  &:after {
+    content: "+";
+    color: white;
```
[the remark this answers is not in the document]
MayukhSamadder2000: As kio mentioned it's better to use a theme variable here instead of hardcoding the color

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  padding: 16px;
+  margin: 5px;
+  border-image: linear-gradient(#346ecd, #cd34a2) 30;
+  border-width: 2px;
+  border-style: solid;
+`;
+
+const Value = styled.div`
+  font-family: 'PTRootUIWebMedium', sans-serif;
+  color: #57c2d6;
```
MayukhSamadder2000, line 87: Better to use color through custom theming variables here.

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  color: #ff0065;
+  display: contents;
+  font-weight: bold;
+`;
+
+const HorizontalLine = styled.div`
+  margin: 9px 0;
+  width: 100%;
+  height: 2px;
+  background: linear-gradient(90deg, #23a9c9, #cd34a2);
```
MayukhSamadder2000, line 102: Better to use color through custom theming variables here.

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  display: flex;
+  color: ${(props) => props.color === 'red' && '#ff0065'};
+`;
+
+const mapOfferToOption = (offer: ExchangeOffer) => {
+  const serviceDetails = swapServiceIdToDetails[offer.provider];
+  return {
+    title: serviceDetails?.title ?? offer.provider,
+    value: offer.provider,
+    iconUrl: serviceDetails?.iconUrl,
```
MayukhSamadder2000, line 123: is there any fallback image for iconUrl?

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+
+    return await Promise.allSettled(accountBalancePromise)
+      .then((response: any[]) => {
+        const accountWalletBalance = response.map(accountBalance => accountBalance?.status == 'fulfilled' && accountBalance?.value);
+
+        var smartWalletBalance = 0;
+        var smartWalletUSD = 0;
+        
+        var keyBasedBalance = 0;
+        var keyBasedUSD = 0;
```
MayukhSamadder2000, line 230: let's try to use `let` if possible over here?

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useEffect, useMemo, useState } from "react";
 import styled, { useTheme } from "styled-components";
 import { ethers } from "ethers";
 import { HiOutlinePencilAlt } from "react-icons/hi";
-import { BsClockHistory, BiCheck, IoClose, FaSignature } from "react-icons/all";
+import { BsClockHistory, BiCheck, IoClose, FaSignature , CgSandClock} from "react-icons/all";
```
MayukhSamadder2000, line 5: its better to import the icons from their respective files like `react-icons/fa` `react-icons/io` etc

MayukhSamadder2000 (changes requested):
Can you check the comments?

src/components/AccountSwitchInput/AccountSwitchInput.tsx
```diff
@@ -4,6 +4,8 @@ import { SelectOption } from '../SelectInput/SelectInput';
 import { DestinationWalletEnum } from '../../enums/wallet.enum';
 
 interface AccountSwitchInputProps {
+  smartWalletTotal?:string;
+  keyBasedWalletTotal?:string;
```
poocart, line 8:
```suggestion
  smartWalletTotal?: string;
  keyBasedWalletTotal?: string;
```

src/components/Button/closeButtonWhite.tsx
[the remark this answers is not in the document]
poocart: +1

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+`;
+
+const Container = styled.div`
+  align-items: center;
+  font-family: 'PTRootUIWebMedium', sans-serif;
+  background: #21002e;
+  color: #fefefe;
+  padding: 16px;
+  margin: 5px;
+  border-image: linear-gradient(#346ecd, #cd34a2) 30;
```
poocart, line 80: @parag-chauhanit might be repeating theme comment, but make sure you go through all of your code to check ANY colours for theme support

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  width: 100%;
+  height: 2px;
+  background: linear-gradient(90deg, #23a9c9, #cd34a2);
+`;
+
+const Bold = styled.p`
+  font-weight: bold;
+`;
+
+//TODO : change color of block if polygon has less than 10,000 plr on keybased / smart wallet
```
poocart, line 109: was this part of ticket scope? yes – please complete, no – remove and add a ticket to backlog

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+const Block = styled.div`
+  font-size: 12px;
+  padding: 2px;
+  text-indent: 2px;
+  display: flex;
+  color: ${(props) => props.color === 'red' && '#ff0065'};
```
poocart, line 115: red is valid color name in css, make sure either to rename prop, just `red` prop name maybe? not sure what's the context, but this is wrong

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  const [selectedFromAsset, setSelectedFromAsset] =
+    useState<IAssetWithBalance | null>(values?.selectedAsset ?? null);
+
+  const [selectedAccountType, setSelectedAccountType] = useState<string>(
+    AccountTypes.Contract
+  );
+
+  const [selectedFromNetwork, setSelectedFromNetwork] = useState<Chain | null>(
+    null
+  );
```
poocart, line 163: please make sure you use consistent line breaking, seems that when it comes to hooks you are switching between break on type vs break on value, while here really line lenght is fine and no break needed at all?
```suggestion
  const [selectedFromNetwork, setSelectedFromNetwork] = useState<Chain | null>(null);
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  const fixed = multiCallData?.fixed ?? false;
+  const defaultCustomReceiverAddress =
+    values?.receiverAddress &&
+    !addressesEqual(providerAddress, values?.receiverAddress) &&
+    !addressesEqual(accountAddress, values?.receiverAddress)
+      ? values.receiverAddress
+      : null;
+  const [customReceiverAddress, setCustomReceiverAddress] = useState<
+    string | null
+  >(defaultCustomReceiverAddress);
```
poocart, line 184: please do not break lines when not needed, thanks
```suggestion
  const [customReceiverAddress, setCustomReceiverAddress] = useState<string | null>(defaultCustomReceiverAddress);
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  const [useCustomAddress, setUseCustomAddress] = useState<boolean>(
+    !!defaultCustomReceiverAddress
+  );
```
poocart, line 187:
```suggestion
  const [useCustomAddress, setUseCustomAddress] = useState<boolean>(!!defaultCustomReceiverAddress);
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  const defaultSelectedReceiveAccountType =
+    (!values?.receiverAddress && values?.accountType === AccountTypes.Key) ||
+    (values?.receiverAddress &&
+      values?.accountType === AccountTypes.Contract &&
+      addressesEqual(providerAddress, values?.receiverAddress))
+      ? AccountTypes.Key
+      : AccountTypes.Contract;
+  
+  const [selectedReceiveAccountType, setSelectedReceiveAccountType] =
+    useState<string>(defaultSelectedReceiveAccountType);
```
poocart, line 198:
```suggestion
  const [selectedReceiveAccountType, setSelectedReceiveAccountType] = useState<string>(defaultSelectedReceiveAccountType);
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+    setTransactionBlockFieldValidationError,
+  } = useTransactionBuilder();
+
+  useEffect(() => {
+    resetTransactionBlockFieldValidationError(transactionBlockId, 'amount');
+    resetTransactionBlockFieldValidationError(transactionBlockId, 'offer');
+    resetTransactionBlockFieldValidationError(
+      transactionBlockId,
+      'toAssetAddress'
+    );
```
poocart, line 212:
```suggestion
    resetTransactionBlockFieldValidationError(transactionBlockId, 'toAssetAddress');
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+  }, [selectedFromNetwork]);
+
+  const getWalletBalance = async (chainId: number, name: string) => {
+    const accountBalancePromise = [];
+    if (accountAddress && providerAddress) {
+      accountBalancePromise.push(getSupportedAssetsWithBalancesForChainId(chainId, true, accountAddress));
+      accountBalancePromise.push(getSupportedAssetsWithBalancesForChainId(chainId, true, providerAddress));
+    }
+
+    return await Promise.allSettled(accountBalancePromise)
```
poocart, line 222: no need since your function is already `async`
```suggestion
    return Promise.allSettled(accountBalancePromise)
```

poocart: also I would replace all this code into following (cleaner and more readable – subjective tho):
```
Promise.all([accountAddress, providerAddress].map(async (address) => {
  await getSupportedAssetsWithBalancesForChainId(chainId, true, address)
  ...
  rest of code inside
  ```

File: src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx [hunk 33]
```diff
@@ -0,0 +1,632 @@
...
+      .then((response: any[]) => {
+        const accountWalletBalance = response.map(accountBalance => accountBalance?.status == 'fulfilled' && accountBalance?.value);
```
poocart, line 224: please replace whole code into `Promise.all` example in the above, would make life easier for all of us, thank you

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
poocart: if you ever use `await` please do not mix if with `then...catch`! even though it works it is considered as mixing 2 types of promise resolving (await wrapper with try...catch VS then...catch)

first wait for your method, then do mapping

```suggestion
    const data = await getBalanceForAllChains();
    const accountBalanceWithSupportedChains = data.map((d) => d.value));
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+    }, 0);
+  
+    const totalKeyBasedPLRTokens = accountBalanceWithSupportedChains?.reduce((accumulator, object) => {
+      return accumulator + object.keyBasedWallet;
+    }, 0);
+
+    const totalSmartWalletPLRTokens = accountBalanceWithSupportedChains?.reduce((accumulator, object) => {
+      return accumulator + object.smartWallet;
+    }, 0);
+
```
poocart, line 301: this already looks like it could have reusable method, please add some, thanks

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+    setAccounts(accountBalanceWithSupportedChains);
+  };
+
+  useEffect(() => {
+    fetchAllAccountBalances();
+  }, []);
+
+  const updateAvailableOffers = useCallback<
+    () => Promise<ExchangeOffer[] | undefined>
+  >(
```
poocart, line 315:
```suggestion
  const updateAvailableOffers = useCallback<() => Promise<ExchangeOffer[] | undefined>>(
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+        setIsNFTMember(true);
+      }
+    });
+    
+  }, [sdk, accountAddress, providerAddress]);
+
+  useEffect(() => {
+    // this will ensure that the old data won't replace the new one
+    let active = true;
+    updateAvailableOffers().then((offers) => {
```
poocart, line 369:
```suggestion
    const update = async () => {
      const offers = await updateAvailableOffers();
      ... rest of code
    }
    update();
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+      if (active && offers) {
+        setAvailableOffers(offers);
+        setIsLoadingAvailableOffers(false);
+        if (!offers.length) return;
+        const bestOffer: any = offers?.find((offer) => offer.provider == 'Lifi');
```
poocart, line 374: please use `lifi` as typed constant or use `bridgeServiceIdToDetails`

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+        const selectedOffer = bestOffer?.provider ? bestOffer : offers[0];
+        setSelectedOffer(mapOfferToOption(selectedOffer));
+      }
+    });
+
+    // hook's clean-up function
```
poocart, line 380: no need sit, this is clear enough
```suggestion
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+    selectedOffer,
+    availableOffers,
+    selectedAccountType,
+    receiverAddress,
+  ]);
+
+  const RenderOption = (option: SelectOption) => {
+    const availableOffer = availableOffers?.find(
+      (offer) => offer.provider === option.value
+    );
```
poocart, line 445:
```suggestion
    const availableOffer = availableOffers?.find((offer) => offer.provider === option.value);
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+            <Text size={16} medium>
+              {valueToReceive} {plrDaoAsset.symbol}
+            </Text>
+          )}
+        </div>
+      </OfferDetails>
+    );
+  };
+
+  return !isNFTMember ? (
```
poocart, line 471: false negative check, it could be just retuning the bottom code on `isNFTMember` and then after check closes u can return the rest of this below

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+    <>
+      <Title>Stake into Pillar DAO</Title>
+      <Container>
+        <Text style={{ fontSize: '16px' }}>
+          To become DAO member, you need to stake <Value>10,000 PLR</Value> tokens on Polygon.
+        </Text>
+        <HorizontalLine></HorizontalLine>
+        {
+          <Text style={{ fontSize: '14px' }}>
+            You have{' '}
```
poocart, line 481:
```suggestion
            You have&nbsp;
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+            {hasEnoughPLR ? (
+              <Value>{`${formatAmountDisplay(totalKeyBasedPLRTokens + totalSmartWalletPLRTokens)}`} PLR</Value>
+            ) : (
+              <Total>{`${formatAmountDisplay(totalKeyBasedPLRTokens + totalSmartWalletPLRTokens)}`} PLR</Total>
```
poocart, line 485: what if u define the format before render? would remove duplicate code

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+            )}
+            {' tokens '}
+            {accounts.length > 0
+              ? `on ${accounts.length == 1 ? `${accounts[0].chainName} chain` : `${accounts.length} chains `} on ${
+                  totalKeyBasedPLRTokens > 0 && totalSmartWalletPLRTokens > 0
+                    ? '2 wallets'
+                    : totalKeyBasedPLRTokens > 0
+                    ? 'Key Based'
+                    : 'Smart Wallet'
```
poocart, line 494: please remove multiple ternary as other component or put in memo, but this cannot stay here

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+            errorMessage={errorMessages?.offer}
+          />
+        )}
+      </>
+    </>
+  ) : (
+    <>
+      <Title>Pillar DAO Membership</Title>
+      <Container>
+        <Text style={{ fontSize: '16px' }}>
```
poocart, line 624: use `size` prop

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+        </div>
+      </OfferDetails>
+    );
+  };
+
+  return !isNFTMember ? (
+    <>
+      <Title>Stake into Pillar DAO</Title>
+      <Container>
+        <Text style={{ fontSize: '16px' }}>
```
poocart, line 475: use `size` prop and on other `Text`

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,632 @@
...
+                    ? '2 wallets'
+                    : totalKeyBasedPLRTokens > 0
+                    ? 'Key Based'
+                    : 'Smart Wallet'
+                }`
+              : ''}
+          </Text>
+        }
+        {'\n'}
+        {(totalKeyBasedPLRTokens >= MAX_PLR_TOKEN_LIMIT || totalSmartWalletPLRTokens >= MAX_PLR_TOKEN_LIMIT
```
poocart, line 500: you already have this defined as `hasEnoughPLR`

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -432,14 +458,35 @@ const ActionPreview = ({
   editButtonDisabled = false,
   showSignButton = false,
   showEditButton = false,
+  isSubmitTrue = false,
```
poocart, line 461: rename it into `isSubmitted` or something, not it's wrong

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -432,14 +458,35 @@ const ActionPreview = ({
...
   setIsTransactionDone,
   showStatus = true,
   showGasAssetSelect = false,
 }: TransactionPreviewInterface) => {
+  const [minutes, setMinutes] = useState(0);
+  const [seconds, setSeconds] = useState(30);
```
poocart, line 467: you can use just `seconds` and use `moment` lib to show, please fix this and rest of related code

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -818,6 +865,192 @@ const ActionPreview = ({
     );
   }
 
+  if (type === TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE) {
```
poocart, line 868: please fix spacing and line breaks wirhin this check

src/providers/TransactionBuilderContextProvider.tsx
```diff
@@ -644,7 +650,117 @@ const TransactionBuilderContextProvider = ({
       setTransactionBlocks([]);
       showAlertModal('Transaction sent');
       setIsSubmitting(false);
-    } else {
+
+    }
+    else if (crossChainActions[0].type == TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE) {
```
poocart, line 655:
```suggestion
    } else if (crossChainActions[0].type == TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE) {
```

poocart: hmm this would mean that this would only work if FIRST action is `PLR_DAO_STAKE`, please check how other blocks are handled

src/providers/TransactionBuilderContextProvider.tsx
```diff
@@ -644,7 +650,117 @@ const TransactionBuilderContextProvider = ({
...
+        setIsSubmitting(false);
+        return;
+      }
+
+      let result: {
+        transactionHash?: string;
+        errorMessage?: string;
+      };
+
+      result = crossChainAction.useWeb3Provider
```
poocart, line 669: please reuse what other blocks use besides `KLIMA_STAKE`, it is wrong example to follow

src/providers/TransactionBuilderContextProvider.tsx
```diff
@@ -718,6 +834,10 @@ const TransactionBuilderContextProvider = ({
     (transactionBlock) => transactionBlock.type === TRANSACTION_BLOCK_TYPE.KLIMA_STAKE
   );
 
+  const hasPlrDaoBlockAdded = transactionBlocks.some(
+    (transactionBlock) => transactionBlock.type === TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE,
+  );
```
poocart, line 839:
```suggestion
  const hasPlrDaoBlockAdded = transactionBlocks.some((transactionBlock) => transactionBlock.type === TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE);
```

src/utils/transaction.ts
poocart: hmmm @IAmKio I think this is good cuz it's method name?

src/utils/transaction.ts
```diff
@@ -314,6 +421,308 @@ export const buildCrossChainAction = async (
     }
   }
 
+  if (transactionBlock.type === TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE
+    && !!transactionBlock?.values?.fromChainId
+    && !!transactionBlock?.values?.fromAsset
```
poocart, line 426: please use ONE pattern how you do multi line logic operators, in one piece of code you put it at back of line, here at front, at front is preferred, thank you
```suggestion
    && !!transactionBlock?.values?.fromChainId
    && !!transactionBlock?.values?.fromAsset
```

src/utils/transaction.ts
```diff
@@ -314,6 +421,308 @@ export const buildCrossChainAction = async (
...
+    && !!transactionBlock?.values?.amount
+    && !!transactionBlock?.values?.offer
+    && !!transactionBlock?.values?.receiverAddress
+    && !!transactionBlock?.values?.toAsset) {
+    if(!transactionBlock?.values?.hasEnoughPLR){
```
poocart, line 431: combine under single please
```suggestion
    && !!transactionBlock?.values?.toAsset) {
    if(!transactionBlock?.values?.hasEnoughPLR){
```

src/utils/transaction.ts
```diff
@@ -314,6 +421,308 @@ export const buildCrossChainAction = async (
...
+      // Swap 
+      try {
+        const {
+          values: {
+            amount,
+            fromChainId: chainId ,
```
poocart, line 437:
```suggestion
            fromChainId: chainId,
```

src/utils/transaction.ts
```diff
@@ -933,7 +1343,7 @@ export const estimateCrossChainAction = async (
 
     try {
       for (const transactionsToSend of crossChainAction.transactions) {
-        const { to, data, value } = transactionsToSend;
+        const { to, data, value, createTimestamp} = transactionsToSend;
```
poocart, line 1346:
```suggestion
        const { to, data, value, createTimestamp } = transactionsToSend;
```

src/utils/transaction.ts
```diff
@@ -998,7 +1409,7 @@ export const estimateCrossChainAction = async (
     //
   }
 
-  return { gasCost, errorMessage, usdPrice, feeAmount };
+  return { gasCost, errorMessage, usdPrice, feeAmount, timeStamp };
```
poocart, line 1412: why `timestamp` is needed here on estimation type?

poocart: you can reuse `createdTimestamp` where you need this, you would even have this prior to estimation, unless you need time specific estimations?

src/types/crossChainAction.ts
```diff
@@ -76,6 +86,15 @@ interface KlimaStakingAction {
   gasCost?: string;
 }
 
+interface PlrStakingAction {
+	type: typeof TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE;
```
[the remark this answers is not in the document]
parag-chauhan: We are using the action type to handle the transaction flow and handle the Preview UI. And for other transaction type is required.

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
parag-chauhan: Yes we did, it is in the scope. Removed comment.

src/utils/transaction.ts
parag-chauhan: We need if condition (`if(!transactionBlock?.values?.hasEnoughPLR)`) to handle swap and staking flow.

IAmKio (changes requested):
Just one more additional query but other than that - looking much better! @poocart and @MayukhSamadder2000 could you re-view when you get a moment. Cheers

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -818,6 +858,185 @@ const ActionPreview = ({
...
+
+    return (
+      <Card
+        title="PLR Dao Staking"
+        marginBottom={20}
+        onCloseButtonClick={onRemove}
+        showCloseButton={showCloseButton}
+        additionalTopButtons={additionalTopButtons}
+      >
+        {hasEnoughPLR && (
```
Yash-Chaubal, line 894: You have another check below after this one where you use `!hasEnoughPLR &&` - instead you could shorten this a bit:
```suggestion
        {hasEnoughPLR ? <TrueComponent /> :  <FalseComponent />}
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,562 @@
...
+  font-weight: bold;
+`;
+
+const HorizontalLine = styled.div`
+  margin: 9px 0;
+  width: 100%;
+  height: 2px;
+  ${({ theme }) => theme.color.text.tokenTotal}
+  background: ${({ theme }) =>
+    `linear-gradient(90deg, ${theme.color.background.horizontalLineFromColor} , ${theme.color.background.horizontalLineToColor})`};
```
MayukhSamadder2000, line 99: I think here instead of keeping two colors we can have a single color variable something like
```
background: ${({ theme }) => theme.color.horizontalLineColor };
```
Reason being in this case if people want they can also add linear gradient , radial gradient and a single color too.

src/utils/chain.ts
```diff
@@ -15,6 +15,7 @@ export const CHAIN_ID = {
   FANTOM: 250,
   CELO: 42220,
   MOONBEAM: 1284,
+  GNOSIS: 100,
```
Yash-Chaubal, line 18: `GNOSIS` is a duplicate of `XDAI` (line 10), could you use that instead?

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,562 @@
...
+  fromAsset?: IAssetWithBalance;
+  selectedAsset?: IAssetWithBalance | null;
+  toAsset?: TokenListToken;
+  fromAssetDecimals?: number;
+  fromAssetSymbol?: string;
+  amount?: string;
+  accountType: AccountTypes;
+  receiverAddress?: string;
+  offer?: ExchangeOffer;
+}
```
MayukhSamadder2000, line 46: can we arrange this in a way so that mandatory types can come at the top and optional ones can come at the bottom?
Just for readability.

src/utils/chain.ts
```diff
@@ -196,6 +197,30 @@ export const klimaAsset: IAssetWithBalance = {
   balanceWorthUsd: null,
 }
 
+export const plrDaoAsset: IAssetWithBalance = {
+  address: '0xa6b37fc85d870711c56fbcb8afe2f8db049ae774',
+  chainId: 137,
```
Yash-Chaubal, line 202: For the `chainId` could you use the constants from above (`CHAIN_ID.POLYGON`)

src/utils/chain.ts
```diff
@@ -196,6 +197,30 @@ export const klimaAsset: IAssetWithBalance = {
...
+  decimals: 18,
+  logoURI: 'https://public.etherspot.io/buidler/chain_logos/polygon.svg',
+  balance: BigNumber.from(0),
+  assetPriceUsd: null,
+  balanceWorthUsd: null,
+}
+
+export const plrDaoMemberNFT = {
+  address: '0xa6b37fc85d870711c56fbcb8afe2f8db049ae774',
+  chainId: 137,
```
Yash-Chaubal, line 214: Same as above

src/utils/transaction.ts
```diff
@@ -38,6 +39,108 @@ import { CROSS_CHAIN_ACTION_STATUS } from '../constants/transactionDispatcherCon
...
+  try {
+    const routes = await sdk.getAdvanceRoutesLiFi({
+      fromChainId,
+      toChainId, //Polygon
+      fromAmount,
+      fromTokenAddress,
+      toTokenAddress,
+      toAddress,
+    });
+    const bestRoute = routes.items.reduce((best: any, route) => {
```
Yash-Chaubal, line 60: Do we know what type `best` would be? It's better to avoid `any` where we can

src/utils/transaction.ts
```diff
@@ -38,6 +39,108 @@ import { CROSS_CHAIN_ACTION_STATUS } from '../constants/transactionDispatcherCon
...
+      if (!best || BigNumber.from(best['toAmount']).lt(route.toAmount)) return route;
```
Yash-Chaubal, line 61: Also this isn't wrong, just a consistency thing - and also added a check to see if `toAmount` exists on `best`, since it's of an `any` type
```suggestion
      if (!best?.toAmount || BigNumber.from(best.toAmount).lt(route.toAmount)) return route;
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,562 @@
...
+
+  const getWalletBalance = async (chainId: number, name: string) => {
+    return Promise.allSettled(
+      [accountAddress, providerAddress].map(async (address) => {
+        return await getSupportedAssetsWithBalancesForChainId(chainId, true, address);
+      })
+    )
+      .then((response: any[]) => {
+        const accountWalletBalance = response.map(
+          (accountBalance) => accountBalance?.status == 'fulfilled' && accountBalance?.value
```
MayukhSamadder2000, line 199: Are we sure we're using map here? Won't the `filter` method be a better choice?

src/utils/transaction.ts
```diff
@@ -314,6 +524,266 @@ export const buildCrossChainAction = async (
...
+        try {
+          const bestRoute = await fetchBestRoute(
+            sdk,
+            fromChainId,
+            plrDaoAsset.chainId,
+            amountBN,
+            fromAssetAddress,
+            plrDaoAsset.address,
+            sdk.state.accountAddress
+          );
```
Yash-Chaubal, line 644: On an error you return an error message from `fetchBestRoute`, but we don't seem to use it. Is this something we need to show this to the user?

src/utils/transaction.ts
```diff
@@ -314,6 +524,266 @@ export const buildCrossChainAction = async (
...
+
+          const { items: advancedRouteSteps } = await sdk.getStepTransaction({ route: bestRoute });
```
Yash-Chaubal, line 646: Also there's no check to see if `bestRoute` is actually returned or not instead of an error before we use it

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,562 @@
...
+            {' tokens '}
+            {accounts.length > 0 ? `on ${chain} on ${wallet}` : ''}
+          </Text>
+        }
+        {'\n'}
+        {tokenArray.map(({ chainName, keyBasedWallet, smartWallet }) => (
+          <Text size={12}>
+            {<Block></Block>}
+            {keyBasedWallet > 0 && (
+              <Block color={chainName === 'Polygon' && keyBasedWallet < MAX_PLR_TOKEN_LIMIT ? '#ff0065' : ''}>
```
MayukhSamadder2000, line 447: let's pass color through theme.

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,562 @@
...
+                {`\u25CF`}
+                <Bold>{formatAmountDisplay(keyBasedWallet)} PLR</Bold> on <Bold>{chainName}</Bold> on{' '}
+                <Bold> Keybased Wallet</Bold>
+              </Block>
+            )}
+            {smartWallet > 0 && (
+              <Block color={chainName === 'Polygon' && smartWallet < MAX_PLR_TOKEN_LIMIT ? '#ff0065' : ''}>
```
MayukhSamadder2000, line 454: let's pass color through theme.

src/utils/transaction.ts
```diff
@@ -314,6 +524,266 @@ export const buildCrossChainAction = async (
...
+              logoURI: fromAssetIconUrl,
+            },
+            amount,
+            offer,
+            accountType,
+          },
+        } = transactionBlock;
+
+        const amountBN = ethers.utils.parseUnits(amount, fromAssetDecimals);
+        try {
```
Yash-Chaubal, line 635: These `try` blocks are pretty large, and ideally we shouldn't be nesting a try inside another try - the first one spans from lines 618-781, and the nested from lines 635-778. 

Could you break these down, and only wrap things that need to be wrapped in try blocks (like interactions with the sdk or a library etc)

MayukhSamadder2000 (changes requested):
Added few more comments

Yash-Chaubal (changes requested):
Good job so far! I've added some more comments just about code formatting and error checking. I still need to go through the other half, but thought I'd post the comments I have so far.

Also for the linter I'm using the default Prettier settings + `.editorconfig` file - just make sure in the Prettier settings, you tick the box that asks if you want to override the settings with that file

src/utils/transaction.ts
parag-chauhan: Not necessarily! think we can avoid it. Will make this change.

src/utils/transaction.ts
parag-chauhan: I guess it will be better to use the above error message over here in case the best route function fails.

src/utils/transaction.ts
parag-chauhan: Using this error message in its parent function in case bestRoute function fails we can return it.

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -432,14 +459,27 @@ const ActionPreview = ({
...
+
+  const { preview, chainId, type, estimated, isEstimating } = crossChainAction;
+
+  useEffect(() => {
+    if (isSubmitted) {
+      let interval = setInterval(() => {
+        setTimer(timer + 1);
+      }, 1000);
+      return () => clearInterval(interval);
+    }
```
poocart, line 480: make sure u always return timeout clear, move this line outside `if` and just add simple condition if interval is set before clearing one

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -818,6 +858,184 @@ const ActionPreview = ({
     );
   }
 
+  if (type === TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE) {
+    const { fromAsset, fromChainId, toAsset, providerName, providerIconUrl, receiverAddress, hasEnoughPLR } = preview;
+
+    const previewList = crossChainAction?.batchTransactions?.length
+      ? crossChainAction?.batchTransactions.map((action) =>
+          action.type === TRANSACTION_BLOCK_TYPE.ASSET_SWAP ||
```
poocart, line 866: isn't this `PLR_DAO_STAKE` block type?

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -818,6 +858,184 @@ const ActionPreview = ({
...
+            ? action.preview
+            : null
+        )
+      : [crossChainAction.preview];
+
+    const fromNetwork = supportedChains.find((supportedChain) => supportedChain.chainId === fromChainId);
+
+    const toNetwork = supportedChains[1];
+
+    const toChainTitle = toNetwork?.title ?? CHAIN_ID_TO_NETWORK_NAME[CHAIN_ID.POLYGON].toUpperCase();
```
poocart, line 877: staking is only on Polygon?

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -818,6 +858,184 @@ const ActionPreview = ({
...
+          </ValueWrapper>
+          <ValueWrapper>
+            {!!cost && (
+              <>
+                <ColoredText>Gas price</ColoredText>
+                {cost}
+              </>
+            )}
+            {crossChainAction.transactions[crossChainAction.transactions.length - 1].createTimestamp && (
+              <>
```
poocart, line 981: maybe create const for `crossChainAction.transactions[crossChainAction.transactions.length - 1].createTimestamp` just before render return

src/utils/chain.ts
```diff
@@ -196,6 +196,30 @@ export const klimaAsset: IAssetWithBalance = {
   balanceWorthUsd: null,
 }
 
+export const plrDaoAsset: IAssetWithBalance = {
```
poocart, line 199: please move these to different utils than `chain` maybe create another one called assets? these are not chain related

src/utils/transaction.ts
```diff
@@ -160,6 +325,113 @@ export const klimaDaoStaking = async (
...
+      createTimestamp,
+      status: CROSS_CHAIN_ACTION_STATUS.UNSENT,
+    };
+
+    transactions = [...transactions, plrDaoApprovalTransaction, plrDaoStakinglTransaction];
+
+    if (receiverAddress && receiverAddress !== sdk.state.accountAddress) {
+      const sPlrDaoTokenAbi = [
+        "function transfer(address to, uint256 value)",
+      ]
```
poocart, line 409:
```suggestion
      ];
```

src/utils/transaction.ts
```diff
@@ -160,6 +325,113 @@ export const klimaDaoStaking = async (
...
+      to: plrDaoApprovalTransactionRequest.to,
+      data: plrDaoApprovalTransactionRequest.data,
+      chainId: plrDaoAsset.chainId,
+      value: 0,
+      createTimestamp,
+      status: CROSS_CHAIN_ACTION_STATUS.UNSENT,
+    };
+
+    const plrDaoStakingAbi = [
+      "function stake(uint256 value)",
```
poocart, line 387:
```suggestion
      'function stake(uint256 value)',
```

src/utils/transaction.ts
```diff
@@ -160,6 +325,113 @@ export const klimaDaoStaking = async (
...
+      value: 0,
+      createTimestamp,
+      status: CROSS_CHAIN_ACTION_STATUS.UNSENT,
+    };
+
+    transactions = [...transactions, plrDaoApprovalTransaction, plrDaoStakinglTransaction];
+
+    if (receiverAddress && receiverAddress !== sdk.state.accountAddress) {
+      const sPlrDaoTokenAbi = [
+        "function transfer(address to, uint256 value)",
```
poocart, line 408:
```suggestion
        'function transfer(address to, uint256 value)',
```

src/utils/transaction.ts
```diff
@@ -314,6 +586,203 @@ export const buildCrossChainAction = async (
...
+            },
+            offer,
+            receiverAddress,
+            accountType,
+          },
+        } = transactionBlock;
+
+        const fromAmountBN = ethers.utils.parseUnits(amount, fromAssetDecimals);
+
+        const swapServiceDetails = swapServiceIdToDetails[offer.provider];
```
poocart, line 625: sorry I don't get it, is it always offer included in plr staking?

src/utils/transaction.ts
```diff
@@ -314,6 +586,203 @@ export const buildCrossChainAction = async (
...
+              logoURI: fromAssetIconUrl,
+            },
+            amount,
+            offer,
+            accountType,
+          },
+        } = transactionBlock;
+
+        const amountBN = ethers.utils.parseUnits(amount, fromAssetDecimals);
+        const routeData = await fetchBestRoute(
```
poocart, line 697: and always bridge? even if I'm staking on destination chain and have enough PLR there?

src/utils/transaction.ts
```diff
@@ -314,6 +586,203 @@ export const buildCrossChainAction = async (
...
+  if (
+    transactionBlock.type === TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE &&
+    !!transactionBlock?.values?.fromChainId &&
+    !!transactionBlock?.values?.fromAsset &&
+    !!transactionBlock?.values?.amount &&
+    !!transactionBlock?.values?.offer &&
+    !!transactionBlock?.values?.receiverAddress &&
+    !!transactionBlock?.values?.toAsset
+  ) {
+    if (!transactionBlock?.values?.hasEnoughPLR) {
```
poocart, line 598: I think I left comment earlier that it does not read well as you're checking negative value first against positive expectation

you can also avoid `else` by just returning from previous `if` and code won't proceed – saves spaces and reads easier

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
...
+import styled, { useTheme } from 'styled-components';
+import { AccountTypes, ExchangeOffer, NftList } from 'etherspot';
+import { TokenListToken } from 'etherspot/dist/sdk/assets/classes/token-list-token';
+import { ethers } from 'ethers';
+import debounce from 'debounce-promise';
+
+// Types
+import { IPlrDaoStakingMembershipBlock } from '../../types/transactionBlock';
+
+// Components
```
poocart, line 11: please keep it consistent on your own code mate, thanks!
```suggestion
// components
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import styled, { useTheme } from 'styled-components';
+import { AccountTypes, ExchangeOffer, NftList } from 'etherspot';
+import { TokenListToken } from 'etherspot/dist/sdk/assets/classes/token-list-token';
+import { ethers } from 'ethers';
+import debounce from 'debounce-promise';
+
+// Types
```
poocart, line 8:
```suggestion
// types
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
...
+import { IAssetWithBalance } from '../../providers/EtherspotContextProvider';
+
+// utils
+import { formatAmountDisplay, formatMaxAmount } from '../../utils/common';
+import { addressesEqual, isValidEthereumAddress, isValidAmount } from '../../utils/validation';
+import { Chain, supportedChains, plrDaoAsset, plrDaoMemberNFT } from '../../utils/chain';
+import { swapServiceIdToDetails } from '../../utils/swap';
+import { Theme } from '../../utils/theme';
+
+//constants
```
poocart, line 30:
```suggestion
// constants
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
...
+  } = useTransactionBuilder();
+
+  useEffect(() => {
+    // Reset transaction block field errors
+    resetTransactionBlockFieldValidationError(transactionBlockId, 'amount');
+    resetTransactionBlockFieldValidationError(transactionBlockId, 'offer');
+    resetTransactionBlockFieldValidationError(transactionBlockId, 'toAssetAddress');
+  }, [selectedFromNetwork]);
+
+  const getWalletBalance = async (chainId: number, name: string) => {
```
poocart, line 192: please rewrite this into:
1. no `then` on `await`
2. Promise.all to resolve

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
...
+      .catch((error) => {
+        return {
+          chainName: name,
+          keyBasedWallet: 0,
+          smartWallet: 0,
+        };
+      });
+  };
+
+  const getBalanceForAllChains = async () => {
```
poocart, line 232: please rewrite this into:

no then on await
catch with try+catch wrapper
Promise.all to resolve

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
...
+    const chainPromise: any[] = [];
+    supportedChains.forEach(async (chain) => {
+      chainPromise.push(getWalletBalance(chain.chainId, chain.title));
+    });
+    return Promise.allSettled(chainPromise).catch((e) => {
+      return [];
+    });
+  };
+
+  const getTotal = (accountBalanceWithSupportedChains: any[], key: any) => {
```
poocart, line 242: leave no `any`

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
...
+      accountBalanceWithSupportedChains = accountBalanceWithSupportedChains?.filter(
+        (data: any) => data.keyBasedWallet > 0 || data.smartWallet
+      );
+      const totalKeyBasedPLRTokens = getTotal(accountBalanceWithSupportedChains, 'keyBasedWallet');
+      const totalSmartWalletPLRTokens = getTotal(accountBalanceWithSupportedChains, 'smartWallet');
+
+      setTotalKeyBasedPLRTokens(totalKeyBasedPLRTokens);
+      setTotalSmartWalletPLRTokens(totalSmartWalletPLRTokens);
+      setAccounts(accountBalanceWithSupportedChains);
+    } catch (e) {}
```
poocart, line 262:
```suggestion
    } catch (e) {
      //
    }
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
...
+  };
+
+  useEffect(() => {
+    // Fetch token balance for all the chains
+    fetchAllAccountBalances();
```
poocart, line 267: you can just add the method inside useEffect and call it within

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
...
+      }
+    }, 200),
+    [sdk, selectedFromAsset, amount, selectedFromNetwork, accountAddress, selectedAccountType]
+  );
+
+  useEffect(() => {
+    // Fetch a list of NFTs for the account to check if the user is existing member of PLR Dao.
+    try {
+      if (accountAddress && providerAddress) {
+        sdk
```
poocart, line 304: this can be just
```suggestion
      if (!accountAddress || providerAddress || !sdk) return;
      sdk.getNftList() part ...
```

poocart: also consider using `await` instead of `then`

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
...
+      try {
+        const offers = await updateAvailableOffers();
+        if (active && offers) {
+          setAvailableOffers(offers);
+          setIsLoadingAvailableOffers(false);
+          if (!offers.length) return;
+          const bestOffer: any = offers?.find((offer) => offer.provider === swapServiceIdToDetails['Lifi'].title);
+          const selectedOffer = bestOffer?.provider ? bestOffer : offers[0];
+          setSelectedOffer(mapOfferToOption(selectedOffer));
+        }
```
poocart, line 331: consistency in your own code
```suggestion
        if (!active || !offers) return;
        setAvailableOffers(offers);
        setIsLoadingAvailableOffers(false);
        if (!offers.length) return;
        const bestOffer: any = offers?.find((offer) => offer.provider === swapServiceIdToDetails['Lifi'].title);
        const selectedOffer = bestOffer?.provider ? bestOffer : offers[0];
        setSelectedOffer(mapOfferToOption(selectedOffer));
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -0,0 +1,571 @@
...
+            account: accountAddress || providerAddress,
+          })
+          .then((output: NftList) => {
+            let hasNFTContractAddress = output?.items?.filter((nft) => nft.contractAddress === plrDaoMemberNFT.address);
+            if (hasNFTContractAddress?.length) {
+              setIsNFTMember(true);
+            }
+          });
+      }
+    } catch (error) {}
```
poocart, line 315:
```suggestion
    } catch (error) {
      //
    }
```

src/utils/transaction.ts
parag-chauhan: Yes offer is required and we are handling offers in the main screen. We need offer object while doing transaction
image

MayukhSamadder2000 (approved):
LGTM! Do check the comments by others before merging.
Thanks awesome work 🎉

src/utils/transaction.ts
```diff
@@ -38,6 +42,168 @@ import { CROSS_CHAIN_ACTION_STATUS } from '../constants/transactionDispatcherCon
...
+  toChainId: number,
+  fromAmount: BigNumber,
+  fromTokenAddress: string,
+  toTokenAddress: string,
+  toAddress: string
+) => {
+  try {
+    const createTimestamp = +new Date();
+    const routes = await sdk.getAdvanceRoutesLiFi({
+      fromChainId,
```
Yash-Chaubal, line 57: This try block is way too large - ideally we shouldn't be wrapping entire functions in a try block, as this could turn into testing hell, but rather just individual components that actually need to tested (such as interactions with an sdk or fetch calls)

Better to have more but smaller try blocks then one large one

src/utils/transaction.ts
```diff
@@ -38,6 +42,168 @@ import { CROSS_CHAIN_ACTION_STATUS } from '../constants/transactionDispatcherCon
...
+  offer: ExchangeOffer | null,
+  receiverAddress?: string,
+  sdk?: EtherspotSdk | null
+): Promise<{ errorMessage?: string; result?: { transactions: ICrossChainActionTransaction[] } }> => {
+  if (!sdk) return { errorMessage: 'No sdk found' };
+
+  if (!offer) {
+    return { errorMessage: 'Failed build PLR Dao Staking transaction!' };
+  }
+  try {
```
Yash-Chaubal, line 149: Same as above

src/utils/transaction.ts
```diff
@@ -160,6 +326,124 @@ export const klimaDaoStaking = async (
...
+  amount: string,
+  receiverAddress?: string,
+  sdk?: EtherspotSdk | null
+): Promise<{
+  errorMessage?: string;
+  result?: { transactions: ICrossChainActionTransaction[]; returnAmount: BigNumber; provider?: ExchangeProviders };
+}> => {
+  if (!sdk) return { errorMessage: 'No sdk found' };
+
+  try {
```
Yash-Chaubal, line 340: Same as above

src/utils/transaction.ts
```diff
@@ -314,6 +598,205 @@ export const buildCrossChainAction = async (
...
+      !transactionBlock?.values?.offer ||
+      !transactionBlock?.values?.receiverAddress ||
+      !transactionBlock?.values?.toAsset)
+  ) {
+    return { errorMessage: 'Failed to build transaction!' };
+  }
+  if (transactionBlock.type === TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE) {
+    if (transactionBlock?.values?.hasEnoughPLR) {
+      // Staking
+      try {
```
Yash-Chaubal, line 615: Same as above

src/utils/transaction.ts
```diff
@@ -38,6 +42,168 @@ import { CROSS_CHAIN_ACTION_STATUS } from '../constants/transactionDispatcherCon
...
+      const approvalTransaction = {
+        to: approvalTransactionRequest.to,
+        data: approvalTransactionRequest.data,
+        chainId,
+        value: 0,
+        createTimestamp,
+        status: CROSS_CHAIN_ACTION_STATUS.UNSENT,
+      };
+
+      transactions = [approvalTransaction, ...transactions];
```
Yash-Chaubal, line 180: For example instead of wrapping the entire function in a try block, I'd break it down into smaller portions like so
```suggestion
let approvalTransactionRequest: [approvalTransactionRequestType];
try {
      const abi = getContractAbi(ContractNames.ERC20Token);
      const erc20Contract = sdk.registerContract<ERC20TokenContract>('erc20Contract', abi, fromAssetAddress);
      approvalTransactionRequest = erc20Contract?.encodeApprove?.(transactions[0].to, fromAmountBN);
} catch (err) {
      return { errorMessage: 'Failed build transfer transaction!' };
}
      if (!approvalTransactionRequest || !approvalTransactionRequest.to) return { errorMessage: 'Failed build bridge approval transaction!' };

      const approvalTransaction = {
        to: approvalTransactionRequest.to,
        data: approvalTransactionRequest.data,
        chainId,
        value: 0,
        createTimestamp,
        status: CROSS_CHAIN_ACTION_STATUS.UNSENT,
      };

      transactions = [approvalTransaction, ...transactions];
```

Yash-Chaubal (approved):
Hey man great work, most of everything LGTM! I've left some comments about `try` blocks, but in this particular case it should be fine - just notes for the future

src/components/TransactionBlock/TransactionBlock.tsx
```diff
@@ -5,17 +5,15 @@ import SendAssetTransactionBlock from './SendAssetTransactionBlock';
...
-    return <AssetBridgeTransactionBlock {...props} />;
-  if (props.type === TRANSACTION_BLOCK_TYPE.ASSET_SWAP)
-    return <AssetSwapTransactionBlock {...props} />;
-  if (props.type === TRANSACTION_BLOCK_TYPE.KLIMA_STAKE)
-    return <KlimaStakingTransactionBlock {...props} />;
+  if (props.type === TRANSACTION_BLOCK_TYPE.SEND_ASSET) return <SendAssetTransactionBlock {...props} />;
+  if (props.type === TRANSACTION_BLOCK_TYPE.ASSET_BRIDGE) return <AssetBridgeTransactionBlock {...props} />;
+  if (props.type === TRANSACTION_BLOCK_TYPE.ASSET_SWAP) return <AssetSwapTransactionBlock {...props} />;
+  if (props.type === TRANSACTION_BLOCK_TYPE.KLIMA_STAKE) return <KlimaStakingTransactionBlock {...props} />;
+  if (props.type === TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE) return <PlrDaoStakingTransactionBlock {...props} />;
```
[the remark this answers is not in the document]
poocart: if not – discard anyway

src/types/crossChainAction.ts
poocart, line 90: padding

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -159,6 +160,31 @@ const Row = styled.div.attrs((props: { center: boolean }) => props)`
...
+  border-radius: 8px;
+  padding: 8px 14px;
+  word-break: break-all;
+  width: 100%;
+`
+const ColoredText = styled.div`
+  color: ${({ theme }) => theme.color.text.reviewLabel};
+  padding: 14px;
+  word-break: initial;
+`
```
poocart, line 187: line break, semicolon

PR #93, opened 2023-02-17
Asset Bridge changes

parag-chauhan:
## Description
BUIDLER: PLR DAO Membership Block
Add asset bridge to PLR DAO block.

## Motivation and Context
This block is aimed at making the process of joining the PLR DAO easy via the BUIDLER component with the implementation of a PLR  DAO Membership Block. This will allow a user to choose their PLR tokens that they may own on any chain, and ultimately stake it into the PLR DAO staking contract and become a member.

## How Has This Been Tested?
Tested on Polygoan Mainnet
If User have enoughPLR Token on Polygon then user sees the Staking screen
If User doen't have enoughPLR Token on Polygon then user sees the Swap screen
If user selects PLR tokens from chains other than polygon then user sees the Asset Bridge screen.

image
image

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
+++ b/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
@@ -54,6 +54,7 @@ export interface IPlrDaoTransactionBlockValues {
 }
 
 interface AccountBalance {
+  chainId: number;
   chainName: string;
   keyBasedWallet: number;
   smartWallet: number;
@@ -180,7 +181,7 @@ const PlrDaoStakingTransactionBlock = ({
     totalKeyBasedPLRTokens >= MAX_PLR_TOKEN_LIMIT || totalSmartWalletPLRTokens >= MAX_PLR_TOKEN_LIMIT;
   const isPolygonAccountWithEnoughPLR = accounts.some(
     (account) =>
-      account.chainName === 'Polygon' &&
+      account.chainId === CHAIN_ID.POLYGON &&
       selectedFromNetwork?.chainId === CHAIN_ID.POLYGON &&
       selectedFromAsset?.symbol === 'PLR' &&
       ((selectedAccountType === DestinationWalletEnum.Contract && account.smartWallet >= MAX_PLR_TOKEN_LIMIT) ||
@@ -283,12 +284,14 @@ const PlrDaoStakingTransactionBlock = ({
         }
       });
       return {
+        chainId,
         chainName: name,
         keyBasedWallet: keyBasedBalance,
         smartWallet: smartWalletBalance,
       };
     } catch (err) {
       return {
+        chainId,
         chainName: name,
         keyBasedWallet: 0,
         smartWallet: 0,
@@ -530,7 +533,8 @@ const PlrDaoStakingTransactionBlock = ({
       ? 'Key Based'
       : 'Smart Wallet';
   const selectedToChain = supportedChains.find((chain) => chain.chainId === CHAIN_ID.POLYGON);
-  const stakingBalance = isPolygonAccountWithEnoughPLR && `${MAX_PLR_TOKEN_LIMIT}`;
+  const stakingBalance = isPolygonAccountWithEnoughPLR ? `${MAX_PLR_TOKEN_LIMIT}` : '';
+  const newStakingBalance = stakingBalance; 
 
   return (
     <>
@@ -549,13 +553,13 @@ const PlrDaoStakingTransactionBlock = ({
           </Text>
         }
         {'\n'}
-        {tokenArray.map(({ chainName, keyBasedWallet, smartWallet }) => (
+        {tokenArray.map(({ chainId, chainName, keyBasedWallet, smartWallet }) => (
           <Text size={12}>
             {<Block></Block>}
             {keyBasedWallet > 0 && (
               <Block
                 color={
-                  chainName === 'Polygon' && keyBasedWallet < MAX_PLR_TOKEN_LIMIT ? theme?.color?.text?.tokenTotal : ''
+                  chainId === CHAIN_ID.POLYGON && keyBasedWallet < MAX_PLR_TOKEN_LIMIT ? theme?.color?.text?.tokenTotal : ''
                 }
               >
                 {`\u25CF`}
@@ -566,7 +570,7 @@ const PlrDaoStakingTransactionBlock = ({
             {smartWallet > 0 && (
               <Block
                 color={
-                  chainName === 'Polygon' && smartWallet < MAX_PLR_TOKEN_LIMIT ? theme?.color?.text?.tokenTotal : ''
+                  chainId === CHAIN_ID.POLYGON && smartWallet < MAX_PLR_TOKEN_LIMIT ? theme?.color?.text?.tokenTotal : ''
                 }
               >
                 {`\u25CF`}
@@ -598,11 +602,7 @@ const PlrDaoStakingTransactionBlock = ({
             resetTransactionBlockFieldValidationError(transactionBlockId, 'fromAssetSymbol');
             resetTransactionBlockFieldValidationError(transactionBlockId, 'fromAssetDecimals');
             setSelectedFromAsset(asset);
-            if (amountBN) {
-              setAmount(stakingBalance || formatMaxAmount(amountBN, asset.decimals));
-            } else {
-              setAmount(stakingBalance || '');
-            }
+            setAmount(amountBN && !newStakingBalance ? formatMaxAmount(amountBN, asset.decimals) : newStakingBalance);
           }}
           onNetworkSelect={(network) => {
             resetTransactionBlockFieldValidationError(transactionBlockId, 'fromChainId');
--- a/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
+++ b/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
@@ -1,9 +1,9 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import styled, { useTheme } from 'styled-components';
 import { AccountTypes, ExchangeOffer } from 'etherspot';
-import { TokenListToken } from 'etherspot/dist/sdk/assets/classes/token-list-token';
 import { ethers } from 'ethers';
 import debounce from 'debounce-promise';
+import { Route } from '@lifi/sdk';
 
 // types
 import { IPlrDaoStakingMembershipBlock } from '../../types/transactionBlock';
@@ -16,6 +16,7 @@ import NetworkAssetSelectInput from '../NetworkAssetSelectInput';
 import TextInput from '../TextInput';
 import SelectInput, { SelectOption } from '../SelectInput/SelectInput';
 import { CombinedRoundedImages, RoundedImage } from '../Image';
+import RouteOption from '../RouteOption';
 
 // providers
 import { IAssetWithBalance } from '../../providers/EtherspotContextProvider';
@@ -27,6 +28,7 @@ import { Chain, supportedChains, plrDaoMemberNFT, CHAIN_ID } from '../../utils/c
 import { plrDaoAsset } from '../../utils/asset';
 import { swapServiceIdToDetails } from '../../utils/swap';
 import { Theme } from '../../utils/theme';
+import { bridgeServiceIdToDetails } from '../../utils/bridge';
 
 //constants
 import { DestinationWalletEnum } from '../../enums/wallet.enum';
@@ -36,7 +38,7 @@ export interface IPlrDaoTransactionBlockValues {
   fromAsset: IAssetWithBalance;
   selectedAsset?: IAssetWithBalance | null;
   offer: ExchangeOffer;
-  toAsset: TokenListToken;
+  toAsset: IAssetWithBalance;
   fromChainId: number;
   fromAssetDecimals?: number;
   fromAssetAddress?: string;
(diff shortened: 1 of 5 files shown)


src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -153,15 +169,24 @@ const PlrDaoStakingTransactionBlock = ({
...
   const [accounts, setAccounts] = useState<AccountBalance[]>([]);
   const [isNFTMember, setIsNFTMember] = useState<boolean>(false);
 
   const hasEnoughPLR =
     totalKeyBasedPLRTokens >= MAX_PLR_TOKEN_LIMIT || totalSmartWalletPLRTokens >= MAX_PLR_TOKEN_LIMIT;
-
-  const filteredSupportedChains = supportedChains.filter((chain) =>
-    [CHAIN_ID.ETHEREUM_MAINNET, CHAIN_ID.POLYGON, CHAIN_ID.XDAI, CHAIN_ID.BINANCE].includes(chain.chainId)
+  const enoughPolygonPLR = accounts.some((account) =>
+    account.chainName === 'Polygon' && selectedAccountType === 'Contract'
```
poocart, line 182: this would always return check on key based if it's different network

poocart: also it does not check PLR so either rename the variable or check PLR

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -153,15 +169,24 @@ const PlrDaoStakingTransactionBlock = ({
...
+      ? account.smartWallet >= MAX_PLR_TOKEN_LIMIT
+      : account.keyBasedWallet >= MAX_PLR_TOKEN_LIMIT
   );
+  const isPolygonAccountWithEnoughPLR =
```
poocart, line 186: how is this different from `enoughPolygonPLR` (which has comments already too)? imo you should combine this into single

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -426,6 +500,13 @@ const PlrDaoStakingTransactionBlock = ({
     );
   };
 
+  const renderRoutes = (option: SelectOption) => (
```
poocart, line 503:
```suggestion
  const renderRoute = (option: SelectOption) => (
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -518,13 +596,14 @@ const PlrDaoStakingTransactionBlock = ({
             resetTransactionBlockFieldValidationError(transactionBlockId, 'fromAssetSymbol');
             resetTransactionBlockFieldValidationError(transactionBlockId, 'fromAssetDecimals');
             setSelectedFromAsset(asset);
-            setAmount(amountBN ? formatMaxAmount(amountBN, asset.decimals) : '');
+            setAmount(
+              enoughPolygonPLR ? `${MAX_PLR_TOKEN_LIMIT}` : amountBN ? formatMaxAmount(amountBN, asset.decimals) : ''
```
poocart, line 600: do not smash multiple ternary, please break this down into variables or else

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -200,6 +226,43 @@ const PlrDaoStakingTransactionBlock = ({
...
+        const { items: routes } = await sdk.getAdvanceRoutesLiFi({
+          fromChainId: selectedFromNetwork.chainId,
+          toChainId: plrDaoAsset.chainId,
+          fromAmount: ethers.utils.parseUnits(amount, selectedFromAsset.decimals),
+          fromTokenAddress: selectedFromAsset.address,
+          toTokenAddress: plrDaoAsset.address,
+          toAddress: receiverAddress ?? undefined,
+        });
+        setAvailableRoutes(routes);
+        if (routes.length === 1) setSelectedRoute(mapRouteToOption(routes[0]));
```
MayukhSamadder2000, line 252: One question in case we don't have the routes length as 1 how's the scenario handled?

src/utils/transaction.ts
```diff
@@ -716,6 +762,68 @@ export const buildCrossChainAction = async (
...
+            symbol: route.toToken.symbol,
+            amount: route.toAmount,
+            iconUrl: toAssetIconUrl,
+            usdPrice: toAssetUsdPrice ?? undefined,
+          },
+          receiverAddress: transactionBlock?.values?.receiverAddress,
+          route,
+        };
+        const result = await bridgeAssetToPlr(fromChainId, route, sdk);
+        if (result?.errorMessage) return { errorMessage: result.errorMessage };
```
MayukhSamadder2000, line 807: One line gap from top and bottom would be good for readability

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -126,6 +132,16 @@ const mapOfferToOption = (offer: ExchangeOffer) => {
   };
 };
 
+const mapRouteToOption = (route: Route) => {
+  const [fistStep] = route.steps;
+  const serviceDetails = bridgeServiceIdToDetails[fistStep?.toolDetails?.key ?? 'lifi'];
+  return {
+    title: fistStep?.toolDetails?.name ?? serviceDetails?.title ?? 'LiFi',
+    value: route.id,
+    iconUrl: fistStep?.toolDetails?.logoURI ?? serviceDetails?.iconUrl,
```
MayukhSamadder2000, line 141: So this condition here, tell me if I'm mistaken, since the first part can be undefined ie `fistStep?.toolDetails?.logoURI` it also seems the second part can be undefined too ie `serviceDetails?.iconUrl` so are we sure that if first part doesn't exist then the second part will definitely exist?

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -365,6 +431,8 @@ const PlrDaoStakingTransactionBlock = ({
   useEffect(() => {
     // update transaction block with best offer (LiFi)
     const offer = availableOffers?.find((availableOffer) => availableOffer.provider === selectedOffer?.value);
+    const route = availableRoutes?.find((availableRoute) => availableRoute.id === selectedRoute?.value);
```
MayukhSamadder2000, line 434: One more point, not that important since `?.` operator does a lot of checks it's better to avoid it , in some cases. I'm attaching an example response here
Screenshot 2023-02-20 at 8 58 02 PM

MayukhSamadder2000: It's not that important though

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -153,15 +169,25 @@ const PlrDaoStakingTransactionBlock = ({
...
     totalKeyBasedPLRTokens >= MAX_PLR_TOKEN_LIMIT || totalSmartWalletPLRTokens >= MAX_PLR_TOKEN_LIMIT;
-
-  const filteredSupportedChains = supportedChains.filter((chain) =>
-    [CHAIN_ID.ETHEREUM_MAINNET, CHAIN_ID.POLYGON, CHAIN_ID.XDAI, CHAIN_ID.BINANCE].includes(chain.chainId)
+  const isPolygonAccountWithEnoughPLR = accounts.some(
+    (account) =>
+      account.chainName === 'Polygon' &&
+      selectedFromNetwork?.chainId === CHAIN_ID.POLYGON &&
+      selectedFromAsset?.symbol === 'PLR' &&
+      ((selectedAccountType === 'Contract' && account.smartWallet >= MAX_PLR_TOKEN_LIMIT) ||
```
poocart, line 186: there's enum for Contract and Key wallet constants

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -518,13 +598,14 @@ const PlrDaoStakingTransactionBlock = ({
             resetTransactionBlockFieldValidationError(transactionBlockId, 'fromAssetSymbol');
             resetTransactionBlockFieldValidationError(transactionBlockId, 'fromAssetDecimals');
             setSelectedFromAsset(asset);
-            setAmount(amountBN ? formatMaxAmount(amountBN, asset.decimals) : '');
+            if (amountBN) {
+              setAmount(stakingBalance || formatMaxAmount(amountBN, asset.decimals));
+            }
```
poocart, line 603: it should reset amount in case there's no  `amountBN`

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
parag-chauhan: @MayukhSamadder2000  serviceDetails is derived from an object constant **bridgeServiceIdToDetails** like this:
`const serviceDetails = bridgeServiceIdToDetails[fistStep?.toolDetails?.key ?? 'lifi'];
`
So serviceDetails will always be there taking up 'lifi' as the default key if the first value is undefined.

parag-chauhan:
@poocart @MayukhSamadder2000 has reolved the issues on this PR. Please have a look.

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -153,15 +169,25 @@ const PlrDaoStakingTransactionBlock = ({
...
   const [isNFTMember, setIsNFTMember] = useState<boolean>(false);
 
   const hasEnoughPLR =
     totalKeyBasedPLRTokens >= MAX_PLR_TOKEN_LIMIT || totalSmartWalletPLRTokens >= MAX_PLR_TOKEN_LIMIT;
-
-  const filteredSupportedChains = supportedChains.filter((chain) =>
-    [CHAIN_ID.ETHEREUM_MAINNET, CHAIN_ID.POLYGON, CHAIN_ID.XDAI, CHAIN_ID.BINANCE].includes(chain.chainId)
+  const isPolygonAccountWithEnoughPLR = accounts.some(
+    (account) =>
+      account.chainName === 'Polygon' &&
```
poocart, line 183: I know this is valid in code, but just always remember that checking string that is a title or a name is very bad pattern that might lead to issues if chain name changes (probably most of top chains except Ethereum had their names changed in last years), this is ok unless there are no other options, but here we can have plenty of options such as adding chainId instead of name to account and then pulling chain title later when we need it in UI, but here we can check against chain ID then

hope it makes sense

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -518,13 +598,16 @@ const PlrDaoStakingTransactionBlock = ({
             resetTransactionBlockFieldValidationError(transactionBlockId, 'fromAssetSymbol');
             resetTransactionBlockFieldValidationError(transactionBlockId, 'fromAssetDecimals');
             setSelectedFromAsset(asset);
-            setAmount(amountBN ? formatMaxAmount(amountBN, asset.decimals) : '');
+            if (amountBN) {
+              setAmount(stakingBalance || formatMaxAmount(amountBN, asset.decimals));
+            } else {
+              setAmount(stakingBalance || '');
+            }
```
poocart, line 605:
```suggestion
            const newStakingBalance = stakingBalance || ''; 
            setAmount(
              amountBN && !newStakingBalance
                ?  formatMaxAmount(amountBN, asset.decimals))
                :   newStakingBalance
            );
```

src/utils/transaction.ts
```diff
@@ -716,6 +762,70 @@ export const buildCrossChainAction = async (
...
+          destinationCrossChainAction: [],
+          isEstimating: false,
+          estimated: null,
+          useWeb3Provider: accountType === AccountTypes.Key,
+          multiCallData: transactionBlock?.multiCallData,
+        };
+
+        return { crossChainAction };
+      } catch (e) {
+        return { errorMessage: 'Failed to build Asset Bridge transaction!' };
```
poocart, line 826: or similar
```suggestion
        return { errorMessage: 'Failed to build PLR DAO Stake transaction!' };
```

src/utils/transaction.ts
```diff
@@ -204,6 +204,51 @@ const fetchSwapAssetTransaction = async (
   }
 };
 
+const bridgeAssetToPlr = async (chainId: number, route: Route | null, sdk?: EtherspotSdk | null) => {
```
poocart, line 207: add return type and maybe rename as `buildBridgeAssetToPlrTransactions` as this is not actually bridging, but returning some result which is transactions

parag-chauhan:
@poocart made the suggested changes. Can you please have a look?

Yash-Chaubal (approved):
LGTM! Just left a comment to confirm something

MayukhSamadder2000 (approved):
LGTM!

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -622,6 +705,29 @@ const PlrDaoStakingTransactionBlock = ({
             errorMessage={errorMessages?.offer}
           />
         )}
+        {!isPolygonAccountWithEnoughPLR &&
+          enableAssetBridge &&
+          !!selectedFromAsset &&
+          !!amount &&
+          (remainingSelectedFromAssetBalance ?? 0) >= 0 && (
```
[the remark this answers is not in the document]
parag-chauhan: No, this condition would not always be true. The value of remainingSelectedFromAssetBalance could be negative if the amount is larger than the selectedFromAsset balance.

PR #94, opened 2023-02-20
Add deploy account feature

parag-chauhan:
## Description
 Deployments Settings Item

## Motivation and Context
With the execption of Polygon and Gnosis, accounts on supported Etherspot chains are not deployed automatically. This modal screen will allow users to invoke a manual deployment of their account on their respective chain.

## How Has This Been Tested?
Tested on Polygon mainnet.
List  of all deployed and un-deployed chains are displayed to the user.
If the user clicks on any un-deployed chain a confirmation pop-up appears.
If the user clicks on deploy button from confirmation pop-up the chain will be deployed.

image
image

## Types of changes

- [X] New feature (non-breaking change which adds functionality)


parag-chauhan:
@poocart @yash @mayukh @vignesh can you please review this PR.

src/components/Button/SecondaryButton.tsx
```diff
@@ -5,13 +5,17 @@ const SecondaryButton = styled.button<{
...
+  display?: string;
 }>`
   font-family: "PTRootUIWebMedium", sans-serif;
   border: none;
   background: transparent;
   font-size: 16px;
   padding: 0;
   ${({ marginTop }) => marginTop && `margin-top: ${marginTop}px`};
+  ${({ margin }) => margin && `margin: ${margin}`};
+  ${({ display }) => display && `display:${display}`};
```
poocart, line 18:
```suggestion
  ${({ display }) => display && `display: ${display};`}
```

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
+import React, { useState, useEffect } from 'react';
+import styled, { useTheme } from 'styled-components';
+
+//icons
+import { IoMdCheckmark } from 'react-icons/io';
+import { CgSandClock } from 'react-icons/cg';
+
+//components
```
poocart, line 8:
```suggestion
// components
```

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
...
+import Card from '../Card/Card';
+import { RoundedImage } from '../Image';
+import { useEtherspot } from '../../hooks';
+import Modal from '../Modal/Modal';
+import { Paragraph } from '../Text';
+import { PrimaryButton, SecondaryButton } from '../Button';
+import ErrorMessage from '../Error/ErrorMessage';
+
+//utils
```
poocart, line 17:
```suggestion
// utils
```

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
...
+  status: string;
+  value?: any;
+  title?: string;
+  address?: string;
+  createdAt?: Date;
+  state?: string;
+  store?: string;
+  type?: string;
+  updatedAt?: Date;
+}
```
poocart, line 81: please review this interface as some values like `store` are not even used plus `state` has an interface from Etherspot types

src/components/Modal/Modal.tsx
```diff
@@ -10,6 +10,7 @@ const ModalOverlay = styled.div<{ noBackground?: boolean }>`
   padding: 15px;
   background: rgba(0, 0, 0, 0.4);
   border-radius: 12px;
+  z-index: ${({ zIndex }) => zIndex && '1'};
```
poocart, line 13:
```suggestion
  ${({ zIndex }) => zIndex && 'z-index: 1;'}
```

poocart: I would also let `zIndex` to be number or if you want to leave as boolean then change prop to something that would match scenario like `isComponentOnTop` or else rather than value name, reason because you can see how other params that are matching css are accepting same values as css

src/components/Error/ErrorMessage.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface Props {
+  errorMessage: string;
+  onClose?: () => void;
+}
+
+const ErrorMessage: React.FC<Props> = ({ errorMessage, onClose }) => {
+  return (
+    <ErrorBanner>
+      <Error>{errorMessage ?? errorMessage}</Error>
```
poocart, line 13:
```suggestion
      <Error>{errorMessage}</Error>
```

src/components/Error/ErrorMessage.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <CloseButton onClick={onClose} top={0} right={0} />
+    </ErrorBanner>
+  );
+};
+
+const ErrorBanner = styled.div`
+  position: fixed;
```
poocart, line 20: it should stay within component boundaries and not leave outside because component should not affect dapp interface

poocart: also just noticed you are setting it as `relative` below, make sure u fix this

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+const Deployment = () => {
+  const [deployedChains, setDeployedChains] = useState<Account[]>([]);
+  const [undeployedChains, setUndeployedChains] = useState<Account[]>([]);
```
poocart, line 85: param is deployed chains, but type is account? make sure u name this properly

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
...
+  const { getSdkForChainId, accountAddress } = useEtherspot();
+  const theme: Theme = useTheme();
+
+  const fetchDeploymentData = async (chainId: number) => {
+    const sdk = getSdkForChainId(chainId);
+    if (accountAddress && sdk) {
+      try {
+        await sdk.computeContractAccount();
+        let account = await sdk.getAccount();
+        return account;
```
poocart, line 99:
```suggestion
        return sdk.getAccount();
```

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
...
+        supportedChains.map(async ({ chainId, title, iconUrl }) => {
+          const account = await fetchDeploymentData(chainId);
+          if (!account) return;
+          return { ...account, chainId, title, iconUrl };
+        })
+      );
+      const accounts = result.filter((acc) => acc);
+      return Promise.allSettled(accounts).catch(() => {
+        return [];
+      });
```
poocart, line 118: remove catch here as it won't even work because you have higher exception handler set

poocart, line 118: remove catch here as it won't even work because you have higher exception handler set, also please do not use 
```suggestion
      return Promise.all(accounts);
```

poocart, line 118: accounts are not a promises, also what are you filtering here? accounts that are present?

```suggestion
      // filter out undefined
      return result.filter((acc) => !!acc);
```

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
...
+      return [];
+    }
+  };
+
+  const hideModal = () => {
+    setConfirmModal(false);
+  };
+
+  const deploymentData = async () => {
+    const data: Account[] | undefined = await getDeployementStatus();
```
poocart, line 129: it does not return undefined as you: 1) filter out undefined accs, 2) in case if err it returns empt array
```suggestion
    const data: Account[] = await getDeployementStatus();
```

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
...
+    if (!data) return;
```
poocart, line 130: it does not return undefined as you: 1) filter out undefined accs, 2) in case if err it returns empt array

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
...
+    const [unDeployed, deployed] = data.reduce(
+      (acc: Account[][], curr: Account) => {
+        if (curr?.value?.state === 'UnDeployed') {
```
poocart, line 133: make sure u check type using Etherspot lib types

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
...
+      [[], []]
+    );
+    setUndeployedChains(unDeployed);
+    setDeployedChains(deployed);
+  };
+
+  const deploy = async () => {
+    if (!selectedChain || !accountAddress) return;
+    const sdk = getSdkForChainId(selectedChain);
+    if (!sdk) return;
```
poocart, line 149: let's show according errors here as well with `setErrorMessage`

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,232 @@
...
+  useEffect(() => {
+    deploymentData();
+  }, []);
+
+  return (
+    <Card title="Deployments" marginBottom={20}>
+      {errorMessage && <ErrorMessage errorMessage={errorMessage} onClose={handleClose} />}
+      {!!confirmModal && (
+        <Modal zIndex>
+          <Paragraph>Deploy your wallet on Etherium to sign messages</Paragraph>
```
poocart, line 171: 1. Ethereum is the network
2. Ether is the currency

always remember ✌️ 

```suggestion
          <Paragraph>Deploy your wallet on Ethereum to sign messages</Paragraph>
```

src/components/Deployment/Deployment.tsx
parag-chauhan: The purpose of filtering out accounts in this function is to remove any undefined or null values from the result array before passing it to Promise.allSettled() for further processing.

parag-chauhan:
Hi! @poocart  made above changes. Can you please have a look?

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+  font-weight: bold;
+  right: 28px;
+  border: none;
+  cursor: pointer;
+  border-radius: 10px;
+  height: 30px;
+  width: 80px;
+`;
+
+interface DeployChain {
```
Yash-Chaubal, line 72: Maybe call this something like `IDeployChain` or `DeployChainProps` just to signify it's an interface

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+  const fetchDeploymentData = async (chainId: number) => {
+    const sdk = getSdkForChainId(chainId);
+    if (accountAddress && sdk) {
+      try {
+        await sdk.computeContractAccount();
+        return await sdk.getAccount();
+      } catch (err) {
+        //
+      }
+    }
```
Yash-Chaubal, line 100: Just a preference so that there are fewer indents
```suggestion
    if (!accountAddress || !sdk) return;
    
      try {
        await sdk.computeContractAccount();
        return await sdk.getAccount();
      } catch (err) {
        //
      }
```

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+  const [deployedChains, setDeployedChains] = useState<DeployChain[]>([]);
+  const [undeployedChains, setUndeployedChains] = useState<DeployChain[]>([]);
+  const [confirmModal, setConfirmModal] = useState<boolean>(false);
+  const [selectedChain, setSelectedChain] = useState<number | null>(null);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+
+  const { getSdkForChainId, accountAddress } = useEtherspot();
+  const theme: Theme = useTheme();
+
+  const fetchDeploymentData = async (chainId: number) => {
```
Yash-Chaubal, line 91: Also better practice to include what type you're returning (unless it's nothing)
```suggestion
  const fetchDeploymentData = async (chainId: number): Promise<TYPE> => {
```

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+      try {
+        await sdk.computeContractAccount();
+        return await sdk.getAccount();
+      } catch (err) {
+        //
+      }
+    }
+  };
+
+  const getDeploymentStatus = async () => {
```
Yash-Chaubal, line 103: Same as above, include the type being returned as in this case the IDE will think what's being returned is `Promise<{...account, chainId: string, title: string, iconUrl: string}[] | any[]>`, which isn't really good for readability

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+    try {
+      const result = [];
```
Yash-Chaubal, line 105: Type this as well, currently `any[]`

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+  };
+
+  const hideModal = () => {
+    setConfirmModal(false);
+  };
+
+  const deploymentData = async () => {
+    const data: DeployChain[] = await getDeploymentStatus();
+    if (!data) return;
+    const [unDeployed, deployed] = data.reduce(
```
Yash-Chaubal, line 125: `undeployed` is one word, but i can see below that for some reason they treat it as two in the etherspot enum `AccountStates.UnDeployed`

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+      }
+      return result;
+    } catch (err) {
+      return [];
+    }
+  };
+
+  const hideModal = () => {
+    setConfirmModal(false);
+  };
```
Yash-Chaubal, line 120: A smaller one, I think you should move this below to where the function `handleClose` is so that relevant functions are next to each other - right now it's sandwiched by deploy code

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  const deploymentData = async () => {
```
Yash-Chaubal, line 122: Maybe change the name of this function to better describe what it does, since it kinda sounds like a variable. Maybe to `getDeploymentData` or `setDeploymentData`?

src/components/Error/ErrorMessage.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import styled from 'styled-components';
+import CloseButton from '../Button/CloseButton';
+
+interface Props {
```
Yash-Chaubal, line 5: Bit generic, specify what it's for

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+import { Theme } from '../../utils/theme';
+
+const Wrapper = styled.div`
+  margin-top: 10px;
+`;
+
+const Section = styled.div`
+  display: flex;
+  background: ${({ theme }) => theme.color.background.topMenu};
+  margin-top: 5px;
```
Yash-Chaubal, line 30: So I prefer keeping styles at the bottom, but I've noticed you've done it different here and in `ErrorMessage.tsx`, better to choose one way and keep it consistent

Yash-Chaubal (changes requested):
Great work man! I've just added a few comments about typing a few functions abd consistency things

src/components/Button/PrimaryButton.tsx
```diff
@@ -5,14 +5,19 @@ const PrimaryButton = styled.div<{
...
   background: ${({ background, theme }) => background || theme.color.background.button};
   padding: 23px;
   border-radius: 16px;
   ${({ marginTop }) => !!marginTop && `margin-top: ${marginTop}px;`}
   ${({ marginBottom }) => !!marginBottom && `margin-bottom: ${marginBottom}px;`}
+  ${({ display }) => !!display && `display: ${display};`}
+  ${({ justifyContent }) => !!justifyContent && `justify-content: ${justifyContent};`}
   font-size: 20px;
-  color: ${({ theme }) => theme.color.text.button};
+  color: ${({ color, theme }) => color || theme.color.text.button};
```
MayukhSamadder2000, line 20: I think `color **??** theme.color.text.button` would work here as if I am not mistaken the only check needed here is if color is undefined or not right? We may not check falsy.

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+            background={theme?.color?.background?.selectInputScrollbar}
+            color={theme?.color?.text?.listItemQuickButtonPrimary}
+            display="flex"
+            justifyContent="center"
+            onClick={deploy}
+          >
+            Deploy
+          </PrimaryButton>
+          <br />
+          <SecondaryButton margin="0 auto" display="flex" color={theme.color?.text?.card} onClick={hideModal}>
```
MayukhSamadder2000, line 179: if this is component specific we might Just create a Wrapper Styled component that Styles the `SecondaryButton` component instead of passing all these properties. Just a suggestion

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,229 @@
...
+      },
+      [[], []]
+    );
+    setUndeployedChains(unDeployed);
+    setDeployedChains(deployed);
+  };
+
+  const deploy = async () => {
+    if (!selectedChain || !accountAddress) return;
+    const sdk = getSdkForChainId(selectedChain);
```
MayukhSamadder2000, line 142:
```suggestion
    if (!selectedChain || !accountAddress) return;
    
    const sdk = getSdkForChainId(selectedChain);
```

src/components/Error/ErrorMessage.tsx
MayukhSamadder2000: Should be something like `ErrorMessageProps`

src/components/Error/ErrorMessage.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <CloseButton onClick={onClose} top={0} right={0} />
+    </ErrorBanner>
+  );
+};
+
+const ErrorBanner = styled.div`
+  position: relative;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
```
MayukhSamadder2000, line 23: Generally position `absolute` is used here in case you wanna centre an element within a container, any specific reason for using `relative`?

src/components/Modal/Modal.tsx
```diff
@@ -10,6 +10,7 @@ const ModalOverlay = styled.div<{ noBackground?: boolean }>`
   padding: 15px;
   background: rgba(0, 0, 0, 0.4);
   border-radius: 12px;
+  z-index: ${({ isComponentOnTop }) => !!isComponentOnTop && '1'};
```
MayukhSamadder2000, line 13: what happens if you don't pass the `isComponentOnTop` property? `zindex: ;`

MayukhSamadder2000 (changes requested):
Looks good just a couple of comments from my side

src/components/Modal/Modal.tsx
parag-chauhan: Some UI issues with displaying the AuthChain label on top. But realised with some changes it should work without z-index.

Yash-Chaubal (approved):
Thanks for making the changes, LGTM!

parag-chauhan:
@poocart @MayukhSamadder2000 @vignesha22 Can you please have a look into these changes when you are available?

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  const { getSdkForChainId, accountAddress } = useEtherspot();
+  const theme: Theme = useTheme();
+
+  const fetchDeploymentData = async (chainId: number): Promise<Account | undefined> => {
+    const sdk = getSdkForChainId(chainId);
+    if (!accountAddress || !sdk) return;
+    try {
+      await sdk.computeContractAccount();
+      return await sdk.getAccount();
```
poocart, line 56:
```suggestion
      await sdk.computeContractAccount();
      return sdk.getAccount();
```

src/components/Deployment/Deployment.tsx
```diff
@@ -0,0 +1,262 @@
...
+        </Body>
+      </Wrapper>
+      <Wrapper>
+        <Header>
+          <CgSandClock size={20} style={{ marginRight: '2px' }} />
+          <Label>Not Deployed</Label>
+        </Header>
+        <Body>
+          {undeployedChains.map(({ chainId, iconUrl, title }) => {
+            return (
```
poocart, line 169: can return just jsx component without `return`
```suggestion
          {undeployedChains.map(({ chainId, iconUrl, title }) => (
```

src/tests/deployment.test.tsx
```diff
@@ -0,0 +1,51 @@
+import axios from 'axios';
+
+const getAccount = async () => {
+  return await axios.get('https://etherspot.pillarproject.io/', {
```
poocart, line 4:
```suggestion
  return axios.get('https://etherspot.pillarproject.io/', {
```

poocart (approved):
left couple more comments, otherwise lgtm

parag-chauhan:
@MayukhSamadder2000 Can you please review this PR?

PR #99, opened 2023-02-24
added usd prices to routes and offers assets

poocart:
## Description

- https://linear.app/pillarproject/issue/PRO-1076/buidler-give-offers-usd-value

## Motivation and Context

- https://linear.app/pillarproject/issue/PRO-1076/buidler-give-offers-usd-value

## How Has This Been Tested?

- Locally.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/src/hooks/useAssetPriceUsd.tsx
+++ b/src/hooks/useAssetPriceUsd.tsx
@@ -1,0 +1,31 @@
+import React, { useEffect, useMemo, useState } from 'react';
+
+// services
+import { getAssetPriceInUsd } from '../services/coingecko';
+
+const useAssetPriceUsd = (
+  chainId?: number,
+  assetAddress?: string,
+): number | null => {
+  const [assetPriceUsd, setAssetPriceUsd] = useState<number | null>(null);
+
+  useEffect(() => {
+    let shouldUpdate = true;
+
+    (async () => {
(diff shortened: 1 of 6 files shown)


src/hooks/useAssetPriceUsd.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const [assetPriceUsd, setAssetPriceUsd] = useState<number | null>(null);
+
+  useEffect(() => {
+    let shouldUpdate = true;
+
+    (async () => {
+      setAssetPriceUsd(null);
+      if (!assetAddress || !chainId) return;
+
+      const priceUsd = await getAssetPriceInUsd(chainId, assetAddress);
```
Yash-Chaubal, line 19: Should we still be using the coingecko api, I'm assuming it's cos we're only pulling data for asset? Can we use the etherspot rates service for this?

poocart: correct, but will check if I can pull native rate using etherspot SDK, thx

poocart: @Yash-Chaubal updated the code to pull asset rates using SDK if it's not native token, also it will always falls back to coingecko if SDK fails

PR #100, opened 2023-02-27
PRO-1111: Inconsistent Results

Yash-Chaubal:
https://linear.app/pillarproject/issue/PRO-1111/etherspot-services-inconsistent-results

## Description
We were hitting the rates limiter very consistently on reloads, so the assets on the wallet block were loading inconsistently.

Fixes:
- We were using `getExchangeSupportedAssets` to get tokens, but were limited to getting 100 tokens per call, meaning we had to make a minimum of 30 calls just to get supported tokens. Changed this to `getTokenListTokens` which gets you all of them at once - but requires us to instantiate all of the sdks for each network.
- Removed a couple of useEffects getting balances to get wallet totals, and absorbed functionality into a new function, which runs when the relevant tx block is selected. As far as I can see everything still seems to work fine.
- The dApp seemed to get instantiated twice because of React.StrictMode, which doubled the calls. I've removed it in the dApp repo.
- We're making quite a few preflight calls at the start (sometimes as many as 10). I'm still not quite sure why these are happening, or even if they're necessary. 

## How Has This Been Tested?
Locally on my machine.

## Types of changes

### DIFF
--- a/src/components/TransactionBlock/AssetSwapTransactionBlock.tsx
+++ b/src/components/TransactionBlock/AssetSwapTransactionBlock.tsx
@@ -154,7 +154,7 @@ const AssetSwapTransactionBlock = ({
 
   useEffect(() => {
     updateWalletBalances();
-  }, []);
+  }, [sdk, accountAddress]);
 
   useEffect(() => {
     // this will ensure that the old data won't replace the new one
--- a/src/components/TransactionBlock/AssetSwapTransactionBlock.tsx
+++ b/src/components/TransactionBlock/AssetSwapTransactionBlock.tsx
@@ -72,16 +72,16 @@ const AssetSwapTransactionBlock = ({
   const [selectedFromAsset, setSelectedFromAsset] = useState<IAssetWithBalance | null>(values?.fromAsset ?? null);
   const [selectedToAsset, setSelectedToAsset] = useState<TokenListToken | null>(values?.toAsset ?? null);
   const [selectedNetwork, setSelectedNetwork] = useState<Chain | null>(values?.chain ?? null);
-  const [selectedOffer, setSelectedOffer] = useState<SelectOption | null>(values?.offer ? mapOfferToOption(values?.offer) : null);
+  const [selectedOffer, setSelectedOffer] = useState<SelectOption | null>(
+    values?.offer ? mapOfferToOption(values?.offer) : null
+  );
   const [availableToAssets, setAvailableToAssets] = useState<TokenListToken[] | null>(null);
   const [availableOffers, setAvailableOffers] = useState<ExchangeOffer[] | null>(values?.offer ? [values.offer] : null);
   const [isLoadingAvailableToAssets, setIsLoadingAvailableToAssets] = useState<boolean>(false);
   const [isLoadingAvailableOffers, setIsLoadingAvailableOffers] = useState<boolean>(false);
   const [showReceiverInput] = useState<boolean>(!!values?.receiverAddress);
   const [receiverAddress, setReceiverAddress] = useState<string>(values?.receiverAddress ?? '');
-  const [selectedAccountType, setSelectedAccountType] = useState<string>(
-    values?.accountType ?? AccountTypes.Contract,
-  );
+  const [selectedAccountType, setSelectedAccountType] = useState<string>(values?.accountType ?? AccountTypes.Contract);
   const fixed = multiCallData?.fixed ?? false;
 
   const { setTransactionBlockValues, resetTransactionBlockFieldValidationError } = useTransactionBuilder();
@@ -92,7 +92,7 @@ const AssetSwapTransactionBlock = ({
     accountAddress,
     providerAddress,
     smartWalletOnly,
-    handleGetBalances
+    updateWalletBalances,
   } = useEtherspot();
   const theme: Theme = useTheme();
 
@@ -102,7 +102,7 @@ const AssetSwapTransactionBlock = ({
       const supportedAssets = await getSupportedAssetsWithBalancesForChainId(
         multiCallData.chain.chainId,
         false,
-        selectedAccountType === AccountTypes.Contract ? accountAddress : providerAddress,
+        selectedAccountType === AccountTypes.Contract ? accountAddress : providerAddress
       );
       const asset = supportedAssets.find((search) => search.address === multiCallData.token?.address);
       setSelectedFromAsset(asset || null);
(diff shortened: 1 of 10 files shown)


src/providers/EtherspotContextProvider.tsx
```diff
@@ -540,22 +533,20 @@ const EtherspotContextProvider = ({
...
-    const handleBalanceGet = async () => {
+  const handleGetBalances = useCallback(
+    async (force?: boolean) => {
       if (!sdk || !accountAddress) return;
-      await getSmartWalletBalancesByChain(accountAddress, supportedChains);
-    };
-    handleBalanceGet();
-  }, [supportedChains, sdk, accountAddress]);
 
-  useEffect(() => {
```
parag-chauhan, line 551: Hey man! Great work here! Just a suggestion about this useEffect call. I feel it can be optimised by
1) Using Promise.all
2) Removing unnecessary conditions from the useEffect dependencies array to avoid triggering unnecessary function calls.
Something like this:
```
useEffect(async () => {
  if (!!force || !smartWalletBalanceByChain.length) {
    await Promise.all([
      getSmartWalletBalancesByChain(accountAddress, supportedChains),
      getKeybasedWalletBalancesPerChain(providerAddress, supportedChains),
    ]);
  }
}, [force, accountAddress, providerAddress]);
```
But overall it looks good to me.

src/providers/EtherspotContextProvider.tsx
```diff
@@ -540,22 +533,20 @@ const EtherspotContextProvider = ({
...
+      if (!!force || !smartWalletBalanceByChain.length)
+        await getSmartWalletBalancesByChain(accountAddress, supportedChains);
+
+      if (!providerAddress) return;
+
+      if (!!force || !keybasedWalletBalanceByChain.length)
+        await getKeybasedWalletBalancesPerChain(providerAddress, supportedChains);
+    },
+    [sdk, accountAddress, providerAddress, smartWalletBalanceByChain, keybasedWalletBalanceByChain]
+  );
```
parag-chauhan, line 549: Hey man! Great work here! The code appears to be well-structured.
I have some small suggestions
1) The function name (handleGetBalances) does not accurately describe what the function does. A better name could be fetchBalances or updateBalances.
2) The function returns undefined. Since it is an async function, it would be better to return a Promise that resolves when the balance have been fetched.

Yash-Chaubal: Yeah agreed that the name could change. This particular function doesn't need anything to be returned - it's just there to load and store data in the Context, not return data to where it's being called. So it returning `undefined` is intended, since that's the default.

src/components/TransactionBlock/AssetSwapTransactionBlock.tsx
```diff
@@ -115,7 +116,9 @@ const AssetSwapTransactionBlock = ({
   const updateAvailableOffers = useCallback<() => Promise<ExchangeOffer[] | undefined>>(
     debounce(async () => {
       // there is a race condition here
-      if (multiCallData && fixed) { return; }
+      if (multiCallData && fixed) {
+        return;
+      }
```
MayukhSamadder2000, line 121:
```suggestion
      if (multiCallData && fixed) return;
```

MayukhSamadder2000 (approved):
LGTM!

src/components/TransactionBlock/AssetSwapTransactionBlock.tsx
```diff
@@ -147,23 +149,28 @@ const AssetSwapTransactionBlock = ({
...
-          if (offers.length === 1) setSelectedOffer(mapOfferToOption(offers[0]))
-          setIsLoadingAvailableOffers(false);
-        }
-      });
+    updateAvailableOffers().then((offers) => {
+      if (active && offers) {
+        setAvailableOffers(offers);
+        if (offers.length === 1) setSelectedOffer(mapOfferToOption(offers[0]));
+        setIsLoadingAvailableOffers(false);
+      }
```
poocart, line 167: I know it's not ur code, but if u making changes please also apply here try/catch handler with await promise

src/components/TransactionBlock/Wallet/WalletTransactionBlock.tsx
```diff
@@ -64,14 +64,8 @@ const WalletTransactionBlock = ({
...
-  const {
-    providerAddress,
-    accountAddress,
-    getSupportedAssetsWithBalancesForChainId,
-    getNftsForChainId,
-    smartWalletBalanceByChain,
-    sdk,
-  } = useEtherspot();
+  const { providerAddress, accountAddress, getSupportedAssetsWithBalancesForChainId, getNftsForChainId, sdk } =
+    useEtherspot();
```
poocart, line 68: not sure if this reads better

src/providers/EtherspotContextProvider.tsx
```diff
@@ -540,22 +533,20 @@ const EtherspotContextProvider = ({
...
-  useEffect(() => {
-    const handleKeybasedBalanceGet = async () => {
-      const isSdkConnected = sdk && accountAddress;
-      if (!isSdkConnected || !providerAddress) return;
-      await getKeybasedWalletBalancesPerChain(providerAddress, supportedChains);
-    };
-    handleKeybasedBalanceGet();
-  }, [supportedChains, sdk, providerAddress, accountAddress]);
+      if (!!force || !smartWalletBalanceByChain.length)
+        await getSmartWalletBalancesByChain(accountAddress, supportedChains);
```
poocart, line 541:
```suggestion
      if (!!force || !smartWalletBalanceByChain.length) {
        await getSmartWalletBalancesByChain(accountAddress, supportedChains);
      }
```

src/providers/EtherspotContextProvider.tsx
```diff
@@ -540,22 +533,20 @@ const EtherspotContextProvider = ({
...
+
+      if (!providerAddress) return;
+
+      if (!!force || !keybasedWalletBalanceByChain.length)
+        await getKeybasedWalletBalancesPerChain(providerAddress, supportedChains);
```
poocart, line 546:
```suggestion
      if (!!force || !keybasedWalletBalanceByChain.length) {
        await getKeybasedWalletBalancesPerChain(providerAddress, supportedChains);
      }
```

src/providers/EtherspotContextProvider.tsx
poocart: sorry, just looks weird

src/providers/EtherspotContextProvider.tsx
poocart: sorry, just looks weird

poocart (approved):

PR #108, opened 2023-03-06
Add DKU staking contract address

parag-chauhan:
## Description
Transfer their assets crosschain to the PLR DAO Contract chain (should it need to be transferred)
Stake the PLR tokens into the PLR DAO smart contract

## Motivation and Context
This block is aimed at making the process of joining the PLR DAO easy via the BUIDLER component with the implementation of a PLR DAO Membership Block. This will allow a user to choose their PLR tokens that they may own on any chain, and ultimately stake it into the PLR DAO staking contract and become a member.

## How Has This Been Tested?
NFT Member display message UI Changes.

image

## Types of changes

### DIFF
--- a/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
+++ b/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
@@ -372,15 +372,17 @@ const PlrDaoStakingTransactionBlock = ({
   );
 
   const getNftList = async () => {
-    if (!accountAddress || !providerAddress || !sdk) return Promise.resolve();
+    if (!accountAddress || !providerAddress || !sdk) {
+      return;
+    }
     try {
-      const nftCollection = await Promise.all(
-        [providerAddress, accountAddress].map(
-          async (address) => await getNftsForChainId(CHAIN_ID.POLYGON, address, true)
-        )
-      ).then(([provider, account]) => provider.concat(account));
+      const [providerAddressNfts, accountAddressNfts] = await Promise.all([
+        getNftsForChainId(CHAIN_ID.POLYGON, providerAddress, true),
+        getNftsForChainId(CHAIN_ID.POLYGON, accountAddress, true),
+      ]);
+      const nftCollection = [...providerAddressNfts, ...accountAddressNfts];
       setIsNFTMember(nftCollection.some((nft) => addressesEqual(nft.contractAddress, plrDaoMemberNft.address)));
-    } catch (e) {
+    } catch (error) {
       //
     }
   };
@@ -524,7 +526,7 @@ const PlrDaoStakingTransactionBlock = ({
   const totalTokens = formatAmountDisplay(totalKeyBasedPLRTokens + totalSmartWalletPLRTokens);
   const tokenArray = hasEnoughPLR && accounts.length == 1 ? [] : accounts;
 
-  if (isNFTMember) {
+  if (!isNFTMember) {
     return (
       <>
         <Title>Pillar DAO Staking</Title>
@@ -730,7 +732,7 @@ const PlrDaoStakingTransactionBlock = ({
         )}
         {enableAssetBridge && !!selectedFromAsset && !!amount && (remainingSelectedFromAssetBalance ?? 0) >= 0 && (
           <SelectInput
-            label={`Route`}
+            label={"Route"}
             options={availableRoutesOptions ?? []}
             isLoading={isLoadingAvailableRoutes}
             selectedOption={selectedRoute}
--- a/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
+++ b/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
@@ -184,12 +184,12 @@ const PlrDaoStakingTransactionBlock = ({
     (account) =>
       account.chainId === CHAIN_ID.POLYGON &&
       selectedFromNetwork?.chainId === CHAIN_ID.POLYGON &&
-      selectedFromAsset?.symbol === 'PLR' &&
+      selectedFromAsset?.symbol === 'DKU' &&
       ((selectedAccountType === DestinationWalletEnum.Contract && account.smartWallet >= MAX_PLR_TOKEN_LIMIT) ||
         (selectedAccountType === DestinationWalletEnum.Key && account.keyBasedWallet >= MAX_PLR_TOKEN_LIMIT))
   );
   const enableAssetBridge = selectedFromNetwork?.chainId !== CHAIN_ID.POLYGON && selectedFromAsset?.symbol === 'PLR';
-  const enableAssetSwap = selectedFromAsset?.symbol !== 'PLR';
+  const enableAssetSwap = selectedFromAsset?.symbol !== 'DKU';
   const toAsset = isPolygonAccountWithEnoughPLR ? plrDaoMemberNFT : plrDaoAsset;
 
   const targetAssetPriceUsd = useAssetPriceUsd(toAsset.chainId, toAsset.address);
--- a/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
+++ b/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
@@ -542,7 +542,9 @@ const PlrDaoStakingTransactionBlock = ({
       ? 'Key Based'
       : 'Smart Wallet';
   const selectedToChain = supportedChains.find((chain) => chain.chainId === CHAIN_ID.POLYGON);
-  const stakingBalance = isPolygonAccountWithEnoughPLR ? `${MAX_PLR_TOKEN_LIMIT}` : '';
+  const decimals = selectedFromAsset?.decimals ?? 18;
+  const updatedAmount = formatAssetAmountInput(`${MAX_PLR_TOKEN_LIMIT}`, decimals);
+  const stakingBalance = isPolygonAccountWithEnoughPLR ? updatedAmount : '';
 
   return (
     <>
--- a/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
+++ b/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
@@ -526,7 +526,7 @@ const PlrDaoStakingTransactionBlock = ({
   const totalTokens = formatAmountDisplay(totalKeyBasedPLRTokens + totalSmartWalletPLRTokens);
   const tokenArray = hasEnoughPLR && accounts.length == 1 ? [] : accounts;
 
-  if (!isNFTMember) {
+  if (isNFTMember) {
     return (
       <>
         <Title>Pillar DAO Staking</Title>
--- a/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
+++ b/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
@@ -532,7 +532,7 @@ const PlrDaoStakingTransactionBlock = ({
   const totalTokens = formatAmountDisplay(totalKeyBasedPLRTokens + totalSmartWalletPLRTokens);
   const tokenArray = hasEnoughPLR && accounts.length == 1 ? [] : accounts;
 
-  if (!isNFTMember) {
+  if (isNFTMember) {
     return (
       <>
         <Title>Pillar DAO Staking</Title>
--- a/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
+++ b/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
@@ -24,7 +24,7 @@ import { IAssetWithBalance } from '../../providers/EtherspotContextProvider';
 // utils
 import { formatAmountDisplay, formatMaxAmount, formatAssetAmountInput } from '../../utils/common';
 import { addressesEqual, isValidEthereumAddress, isValidAmount } from '../../utils/validation';
-import { Chain, supportedChains, plrDaoMemberNFT, CHAIN_ID } from '../../utils/chain';
+import { Chain, supportedChains, plrDaoMemberNft, CHAIN_ID } from '../../utils/chain';
 import { plrDaoAsset, testPlrDaoAsset } from '../../utils/asset';
 import { swapServiceIdToDetails } from '../../utils/swap';
 import { Theme } from '../../utils/theme';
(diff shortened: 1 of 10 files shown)


src/providers/EtherspotContextProvider.tsx
```diff
@@ -220,9 +221,21 @@ const EtherspotContextProvider = ({
...
+            nativeAsset,
+            ...assets,
+            {
+              address: plrDaoAsset.address,
+              chainId: 137,
+              decimals: 18,
+              logoURI: 'https://tokens.1inch.io/0x7fc66500c84a76ad7e9c93437bfc5ac33e2ddae9.png',
+              name: 'Deku',
+              symbol: 'DKU',
+            },
```
poocart, line 237: can you move this to assets and add a comment by marking as "dummy"

src/utils/asset.ts
```diff
@@ -4,10 +4,10 @@ import { CHAIN_ID } from './chain';
 import { PLR_ADDRESS_PER_CHAIN } from '../constants/assetConstants';
 
 export const plrDaoAsset: IAssetWithBalance = {
-  address: '0xa6b37fc85d870711c56fbcb8afe2f8db049ae774',
+  address: '0x6D5862a18C6a169D44d02a8B726a02A5B707B484',
```
poocart, line 7: as per comment in the above you can then leave `plrDaoAsset` as is and just have DKU mapped from separate constant var

src/utils/transaction.ts
```diff
@@ -717,49 +628,30 @@ export const buildCrossChainAction = async (
...
+            amount: '1',
             iconUrl: 'https://public.pillar.fi/files/pillar-dao-member-badge.png',
           },
           receiverAddress: transactionBlock?.values?.receiverAddress,
-          providerName: result.result?.provider
-            ? swapServiceIdToDetails[result.result.provider].title
-            : 'Unknown provider',
-          providerIconUrl: result.result?.provider ? swapServiceIdToDetails[result.result.provider].iconUrl : '',
+          providerName: '',
+          providerIconUrl: '',
```
poocart, line 641: remove?

src/utils/transaction.ts
```diff
@@ -717,49 +628,30 @@ export const buildCrossChainAction = async (
...
           type: TRANSACTION_BLOCK_TYPE.PLR_DAO_STAKE,
           preview,
-          transactions: routeData.transactions || [],
+          transactions,
           isEstimating: false,
           estimated: null,
-          containsSwitchChain: routeData?.bestRoute?.containsSwitchChain,
-          bridgeUsed: routeData?.bestRoute?.steps[0].tool,
-          receiveAmount: routeData?.bestRoute?.toAmount,
+          receiveAmount: '1',
```
poocart, line 652: always 1?

src/utils/transaction.ts
```diff
@@ -671,39 +560,61 @@ export const buildCrossChainAction = async (
               decimals: fromAssetDecimals,
               logoURI: fromAssetIconUrl,
             },
+            toAsset: { decimals: toAssetDecimals },
             amount,
-            offer,
```
poocart, line 675: no longer doing swaps?

src/utils/transaction.ts
parag-chauhan: It is a required field as other blocks are using this fields in the preview.

src/utils/transaction.ts
parag-chauhan: Yes it will be always 1 PLR DAO NFT.

src/utils/transaction.ts
parag-chauhan: In the Staking flow we don't need offers, so removed that.

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -371,13 +381,14 @@ const PlrDaoStakingTransactionBlock = ({
...
-        account: accountAddress || providerAddress,
-      });
-      let hasNFTContractAddress = output?.items?.filter((nft) => nft.contractAddress === plrDaoMemberNFT.address);
-      if (hasNFTContractAddress?.length) {
-        setIsNFTMember(true);
-      }
+      let accountNFTDetails: NftCollection;
+      let providerNFTDetails: NftCollection;
+      [providerNFTDetails] = await getNftsForChainId(CHAIN_ID.POLYGON, providerAddress, true);
+      [accountNFTDetails] = await getNftsForChainId(CHAIN_ID.POLYGON, accountAddress, true);
```
poocart, line 387: can't u just assign here directly to vars?

poocart: also what if there are more NFTs than this?

poocart: it should: run a check on each address if any of holding nfts contain `plrDaoMemberNFT`

src/utils/transaction.ts
poocart: it is not required field by other blocks, please check your block type and apply changes accordingly

src/utils/transaction.ts
poocart: not sure if it's needed then at all here, but if you think it should be kept let's leave it

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
parag-chauhan: @poocart refactored the code for fetching NFTs. Please have a look.

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -369,15 +379,16 @@ const PlrDaoStakingTransactionBlock = ({
...
-        setIsNFTMember(true);
-      }
+      const [providerNFTDetails, accountNFTDetails] = await Promise.all([
+        getNftsForChainId(CHAIN_ID.POLYGON, providerAddress, true),
+        getNftsForChainId(CHAIN_ID.POLYGON, accountAddress, true),
+      ]);
+      const hasMembershipNFT = providerNFTDetails
+        .concat(accountNFTDetails)
+        .some((nft) => addressesEqual(nft.contractAddress, plrDaoMemberNFT.address));
+      setIsNFTMember(hasMembershipNFT);
```
MayukhSamadder2000, line 391: I think we can convert this function to a normal arrow function instead of an async functionand do `Promise.all(...).then(res => ...)`

src/utils/transaction.ts
```diff
@@ -671,39 +560,61 @@ export const buildCrossChainAction = async (
...
+            }
+            const approvalTransaction = {
+              to: approvalTransactionRequest.to,
+              data: approvalTransactionRequest.data,
+              chainId: fromChainId,
+              value: 0,
+              createTimestamp,
+              status: CROSS_CHAIN_ACTION_STATUS.UNSENT,
+            };
+            transactions = [...transactions, approvalTransaction];
```
MayukhSamadder2000, line 589: in line 569 I see `transactions` being a blank array. Are we manipulating it anywhere prior 589? If not then we may not spread transactions here

src/utils/transaction.ts
```diff
@@ -671,39 +560,61 @@ export const buildCrossChainAction = async (
...
             },
+            toAsset: { decimals: toAssetDecimals },
             amount,
-            offer,
+            receiverAddress,
             accountType,
           },
         } = transactionBlock;
-
+        let transactions: any[] = [];
```
MayukhSamadder2000, line 569: we don't need to assert type here I guess

MayukhSamadder2000 (changes requested):
Looks good just a couple of comments

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
poocart: @MayukhSamadder2000 imo everything here unless I don't understand your comment
what maybe can be improved here is loop  `[providerAddress, accountAddress]` for `getNftsForChainId`

src/utils/asset.ts
poocart: sorry @parag-chauhan, but these changes seems to be just same `plrDaoAsset`, but with demo token values?

poocart: my suggestion was to have separate and leave actual token constants as the same as it might be used in other pieces of code so be careful

src/utils/transaction.ts
poocart: no `any[]` please 🙏

src/utils/transaction.ts
poocart: @parag-chauhan this is not fixed

parag-chauhan:
@poocart @MayukhSamadder2000 Made the suggested changes. Can you please have a look?

src/utils/transaction.ts
```diff
@@ -43,6 +42,22 @@ import { ITransactionBlock } from '../types/transactionBlock';
...
 import { PlrV2StakingContract } from '../types/etherspotContracts';
 
+interface IPillarDAO {
+  encodeDeposit(amount: BigNumber): {
+    to: string;
+    data: string;
+  };
+};
+
+interface IPLRTransaction {
```
Yash-Chaubal, line 52: This is a very minor nitpick, but could we stick to camel case even with symbol names 
```suggestion
interface IPlrTransaction {
```

src/utils/transaction.ts
```diff
@@ -43,6 +42,22 @@ import { ITransactionBlock } from '../types/transactionBlock';
 import { PLR_STAKING_ADDRESS_ETHEREUM_MAINNET, POLYGON_USDC_CONTRACT_ADDRESS } from '../constants/assetConstants';
 import { PlrV2StakingContract } from '../types/etherspotContracts';
 
+interface IPillarDAO {
```
Yash-Chaubal, line 45: Same as above
```suggestion
interface IPillarDao {
```

src/utils/chain.ts
```diff
@@ -197,10 +197,10 @@ export const klimaAsset: IAssetWithBalance = {
 }
 
 export const plrDaoMemberNFT = {
```
Yash-Chaubal, line 199: Same as above
```suggestion
export const plrDaoMemberNft = {
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -368,19 +371,23 @@ const PlrDaoStakingTransactionBlock = ({
...
+    Promise.all(nftPromiseArray)
+      .then(([providerNFTDetails, accountNFTDetails]) => {
+        const hasMembershipNFT = providerNFTDetails
+          .concat(accountNFTDetails)
+          .some((nft) => addressesEqual(nft.contractAddress, plrDaoMemberNFT.address));
+        setIsNFTMember(hasMembershipNFT);
+      })
+      .catch((error) => {
+        //
       });
```
Yash-Chaubal, line 390: I'm sure this works fine, but personally I think writing it out this way makes it more clear + we don't need to have a separate promise array + we aren't wrapping the entire thing in a `Promise.all`
```suggestion
    try {
      const nftCollection = await Promise.all(
        [providerAddress, accountAddress].map(
          async (address) => await getNftsForChainId(CHAIN_ID.POLYGON, address, true)
        )
      ).then(([provider, account]) => provider.concat(account));

      setIsNFTMember(nftCollection.some((nft) => addressesEqual(nft.contractAddress, plrDaoMemberNFT.address)));
    } catch (e) {
      //
    }
```

Yash-Chaubal (commented):
Looking good so far! Just added a suggestion for a promise block + some var name changes.

Also are you using a formatter? If not, please could you get Prettier for VSCode and make sure the file formats on save, so that everything is consistent across everyone's IDEs.

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
MayukhSamadder2000: @poocart my point here was why do we need a async function as we're not returning a promise here we can simply use a normal function and maybe just set the values on the then block.
https://betterprogramming.pub/should-i-use-promises-or-async-await-126ab5c98789#:~:text=To%20run%20promises,to%20neglect%20this.

MayukhSamadder2000 (approved):
LGTM!

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -369,16 +372,15 @@ const PlrDaoStakingTransactionBlock = ({
...
-      let hasNFTContractAddress = output?.items?.filter((nft) => nft.contractAddress === plrDaoMemberNFT.address);
-      if (hasNFTContractAddress?.length) {
-        setIsNFTMember(true);
-      }
-    } catch (error) {
+      const nftCollection = await Promise.all(
+        [providerAddress, accountAddress].map(
+          async (address) => await getNftsForChainId(CHAIN_ID.POLYGON, address, true)
+        )
+      ).then(([provider, account]) => provider.concat(account));
```
poocart, line 381: PLEASE PLEASE PLEASE never mix `await` with `then` and never mix `await` with `.catch`

also please if it's a promise function u don't need to return additional await

this is much easier to read and fixes the above:

```suggestion
      const [providerAddressNfts, accountAddressNfts] = await Promise.all([providerAddress, accountAddress].map((address) => getNftsForChainId(CHAIN_ID.POLYGON, address, true)));
      const nftCollection = [...providerAddressNfts, ...accountAddressNfts];
```

src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -717,29 +728,25 @@ const PlrDaoStakingTransactionBlock = ({
...
-              renderSelectedOptionContent={renderRoute}
-              errorMessage={errorMessages?.route}
-              disabled={!availableRoutesOptions?.length || isLoadingAvailableRoutes}
-              noOpen={!!selectedRoute && availableRoutesOptions?.length === 1}
-              forceShow={!!availableRoutesOptions?.length && availableRoutesOptions?.length > 1}
-            />
-          )}
+        {enableAssetBridge && !!selectedFromAsset && !!amount && (remainingSelectedFromAssetBalance ?? 0) >= 0 && (
+          <SelectInput
+            label={`Route`}
```
poocart, line 733:
```suggestion
            label="Route"
```

src/components/TransactionPreview/ActionPreview.tsx
```diff
@@ -912,7 +924,7 @@ const ActionPreview = ({
         showCloseButton={showCloseButton}
         additionalTopButtons={additionalTopButtons}
       >
-        {isPolygonAccountWithEnoughPLR ? (
+        {!enableAssetSwap && !enableAssetBridge ? (
```
poocart, line 927: you can put `!enableAssetSwap && !enableAssetBridge`  as constant as you using it multiple times
x

src/providers/EtherspotContextProvider.tsx
```diff
@@ -220,9 +221,10 @@ const EtherspotContextProvider = ({
       const hasNativeAsset = assets.some(
         (asset) => !asset.address || addressesEqual(asset.address, nativeAssetPerChainId[chainId]?.address)
       );
-
-      supportedAssetsPerChainId[assetsChainId] = hasNativeAsset || !nativeAsset ? assets : [nativeAsset, ...assets];
-
+      // To be added back when DKU token is no longer needed. This is for the DKU testing.
```
poocart, line 224:
```suggestion
      // TODO: to be added back when DKU token is no longer needed. This is for the DKU testing.
```

src/utils/transaction.ts
```diff
@@ -43,6 +42,22 @@ import { ITransactionBlock } from '../types/transactionBlock';
 import { PLR_STAKING_ADDRESS_ETHEREUM_MAINNET, POLYGON_USDC_CONTRACT_ADDRESS } from '../constants/assetConstants';
 import { PlrV2StakingContract } from '../types/etherspotContracts';
 
+interface IPillarDao {
+  encodeDeposit(amount: BigNumber): {
+    to: string;
+    data: string;
+  };
+};
```
poocart, line 50:
```suggestion
}
```

poocart (approved):
few small changes 🙏  otherwise lgtm, thanks!

PR #131, opened 2023-04-19
fix: UI enhancements

MayukhSamadder2000:
## Description
Mainly worked on a couple of tickets made by aldin.

Worked on things like:-
- Hidden second status till Approval is done for tokens.
- Changed hover color & added hover to assets in case of wallet.


Yash-Chaubal (approved):
LGTM! Remember to bump the package version, also not sure why the linter didn't catch my suggested change

PR #144, opened 2023-05-10
fix: asset color issues

MayukhSamadder2000:
Some color related fixes on the loader and removed extra margin


Yash-Chaubal (approved):
LGTM!

PR #146, opened 2023-05-12
fix: updated chain assets

MayukhSamadder2000:
Modified the assets function to conditionally fetch the assets by chain.


parag-chauhan (approved):
LGTM

PR #147, opened 2023-05-12
Fixed dynamic gas fix for klima staking

kaushalrajbacancy:
- Fixed dynamic gas fix for Klima staking
## Description

- Fixed dynamic gas fix for Klima staking in builder

## How Has This Been Tested?

- Doing QA.
-
-

## Types of changes

### DIFF
--- a/src/components/TransactionBlock/KlimaStakingTransactionBlock.tsx
+++ b/src/components/TransactionBlock/KlimaStakingTransactionBlock.tsx
@@ -1,8 +1,8 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { debounce } from 'lodash';
 import styled, { useTheme } from 'styled-components';
-import { AccountTypes, BridgingQuote, CrossChainServiceProvider } from 'etherspot';
-import { BigNumber, ethers } from 'ethers';
+import { AccountStates, AccountTypes, BridgingQuote, CrossChainServiceProvider } from 'etherspot';
+import { BigNumber, BigNumberish, ethers } from 'ethers';
 
 // Types
 import { IKlimaStakingTransactionBlock } from '../../types/transactionBlock';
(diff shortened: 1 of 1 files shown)


MayukhSamadder2000 (approved):
LGTM! Assuming the static numbers used here have already been discussed.

src/components/TransactionBlock/KlimaStakingTransactionBlock.tsx
```diff
@@ -279,10 +280,39 @@ const KlimaStakingTransactionBlock = ({
           return;
         }
 
+        const sdkChain = getSdkForChainId(CHAIN_ID.POLYGON);
+        const gasInfo = await sdkChain?.getGatewayGasInfo();
+
+        let priceUsd = await getAssetPriceInUsd(CHAIN_ID.POLYGON,ethers.constants.AddressZero,sdk);
```
IAmKio, line 286: If either of these `await`'s return an error, what is the value of the error? As anything other than `null` would cause the next `if` statement to still pass...

src/components/TransactionBlock/KlimaStakingTransactionBlock.tsx
```diff
@@ -279,10 +280,39 @@ const KlimaStakingTransactionBlock = ({
...
+        if (!gasInfo || !priceUsd) {
+          setTransactionBlockFieldValidationError(transactionBlockId, 'amount', `No Offer found`);
+          resetRoutes();
+          return;
+        }
+
+        let estimatedGas: BigNumberish = 850000;
+        let currentGasPrice = gasInfo?.fast;
```
IAmKio, line 294: We have a nullable here (`?.`) - what happens when `currentGasPrice` is null on this line?

`let gasFees = currentGasPrice.mul(estimatedGas);`

IAmKio (changes requested):
Just a few questions around possible negative scenarios

src/components/TransactionBlock/KlimaStakingTransactionBlock.tsx
kaushalrajbacancy: Yes. We don't need this **"?"**, as we already validate in the above check.

PR #170, opened 2023-06-26
fix: receiver address fix

MayukhSamadder2000:
fixed the receiver address issue

### DIFF
--- a/src/components/TransactionBlock/HoneySwapLPTransactionBlock.tsx
+++ b/src/components/TransactionBlock/HoneySwapLPTransactionBlock.tsx
@@ -185,7 +185,8 @@ const HoneySwapLPTransactionBlock = ({ id: transactionBlockId, errorMessages, va
       fromAssetIconUrl: selectedFromAsset?.logoURI,
       amount,
       accountType: selectedAccountType,
-      receiverAddress: receiverAddress ?? undefined,
+      receiverAddress:
+        (selectedReceiveAccountType === AccountTypes.Key ? providerAddress : accountAddress) ?? undefined,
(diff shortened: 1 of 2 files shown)


src/utils/transaction.ts
```diff
@@ -423,6 +423,8 @@ export const honeyswapLP = async (
 ) => {
   if (!sdk) return { errorMessage: 'No sdk found' };
 
+  console.log('receiverAddress', receiverAddress);
```
IAmKio, line 426: Was this supposed to be here?

src/components/TransactionBlock/HoneySwapLPTransactionBlock.tsx
```diff
@@ -185,7 +185,8 @@ const HoneySwapLPTransactionBlock = ({ id: transactionBlockId, errorMessages, va
       fromAssetIconUrl: selectedFromAsset?.logoURI,
       amount,
       accountType: selectedAccountType,
-      receiverAddress: receiverAddress ?? undefined,
+      receiverAddress:
+        (selectedReceiveAccountType === AccountTypes.Key ? providerAddress : accountAddress) ?? undefined,
```
IAmKio, line 189: What's the UI result of this being undefined here?

IAmKio (changes requested):
Just a few small points

src/components/TransactionBlock/HoneySwapLPTransactionBlock.tsx
MayukhSamadder2000: Nothing it won't let it pass through then during review

PR #171, opened 2023-06-27
Fixed PLR DAO Balance issue [PRO-1364]

parag-chauhan:
## Description
Fixed PLR DAO Balance issue 

## How Has This Been Tested?
- Local testing with PLR DAO block.
- Tested with Pillar-dao-landing repository by installing package locally

## Screenshots:
image

## Types of changes

### DIFF
--- a/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
+++ b/src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
@@ -213,18 +213,18 @@ const PlrDaoStakingTransactionBlock = ({
   const hasEnoughPLR =
     totalKeyBasedPLRTokens >= MAX_PLR_TOKEN_LIMIT || totalSmartWalletPLRTokens >= MAX_PLR_TOKEN_LIMIT;
 
-  const hasNotEnoughPLRForPolygon = useCallback(() => {
+  const hasEnoughPLRForPolygon = useCallback(() => {
     const selectedChainBalance = accounts.find((acc) => acc.chainId === CHAIN_ID.POLYGON);
     if (!selectedChainBalance) return false;
 
     if (selectedAccountType === AccountTypes.Contract) {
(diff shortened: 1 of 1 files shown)


src/components/TransactionBlock/PlrDaoStakingTransactionBlock.tsx
```diff
@@ -212,13 +212,28 @@ const PlrDaoStakingTransactionBlock = ({
...
+    if (!selectedChainBalance) return false;
+
+    if (selectedAccountType === AccountTypes.Contract) {
+      return selectedChainBalance['smartWallet'] < MAX_PLR_TOKEN_LIMIT;
+    } else {
+      return selectedChainBalance['keyBasedWallet'] < MAX_PLR_TOKEN_LIMIT;
+    }
+  }, [accounts, selectedAccountType, selectedFromNetwork]);
+
+  const hasEnoughPLRInPolygon = !hasNotEnoughPLRForPolygon();
```
Yash-Chaubal, line 227: Coudn't this function just be `hasEnoughPlrForPolygon`? Also in this case if they haven't selected a chain balance then `hasEnoughPlrForPolygon` will be true

Also if we're checking if they have enough PLR, shouldn't we be checking `MIN_PLR_TOKEN_LIMIT` instead of the max amount?
```suggestion
  const hasEnoughPLRForPolygon = useCallback(() => {
    const selectedChainBalance = accounts.find((acc) => acc.chainId === CHAIN_ID.POLYGON);
    if (!selectedChainBalance) return false;

    if (selectedAccountType === AccountTypes.Contract) {
      return selectedChainBalance['smartWallet'] > MAX_PLR_TOKEN_LIMIT;
    } else {
      return selectedChainBalance['keyBasedWallet'] > MAX_PLR_TOKEN_LIMIT;
    }
  }, [accounts, selectedAccountType, selectedFromNetwork]);
```

Yash-Chaubal (commented):
LGTM, just had a question

Yash-Chaubal (approved):
LGTM!

